PR #5, opened 2024-03-08
[Fix] JSON serialization in sync streaming and `OplogEntry` `data` field type

Manrich121:
Changes:

- Fix: `OplogEntry.data` incorrectly typed 
- Add: `JsonUtil.json` to make JSON encoding/decoding config consistent everywhere. 
- Updated JSON config:
  * `encodeDefaults` -> Default values will now be encoded
  * `isLenient` -> Less-strict parsing, in particular unquoted keys and string literals will not throw an error
  * `ignoreUnknownKeys` -> Unknown keys encountered during deserialization will not throw an error

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/utils/Json.kt
+++ b/core/src/commonMain/kotlin/com/powersync/utils/Json.kt
@@ -1,0 +1,15 @@
+package com.powersync.utils
+
+import kotlinx.serialization.json.Json
+
+/**
+ * A global instance of a JSON serializer.
+ */
+object JsonUtil {
+    val json = Json {
+        encodeDefaults = true
+        isLenient = true
+        ignoreUnknownKeys = true
+    }
+}
+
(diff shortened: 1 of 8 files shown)


core/src/commonMain/kotlin/com/powersync/utils/Json.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import kotlinx.serialization.json.Json
+
+/**
+ * A global instance of a JSON serializer.
+ */
+object JsonUtil {
+    val json = Json {
+        encodeDefaults = true
+        isLenient = true
```
rkistner, line 11: Is there a reason for this to be true? As far as I understand it's only "invalid" JSON (unquoted keys and string literals) that would be covered by this, and we should not have invalid JSON like that anywhere.

Manrich121: The reason isLenient was used was because there was an unquoted number value the JSON string in data here https://github.com/powersync-ja/powersync-kotlin/blob/main/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt#L357

rkistner: That would have caused Kotlin to treat it as a String instead of a number, which is a bug, and it's better to have a hard error for cases like that. That specific issue should be solved by the changes here (having `data` as a String instead of `Map<String, String>`).

rkistner (approved):
Confirmed fix in a test app

PR #8, opened 2024-03-19
[Fix] Watched queries + Schema indexes

stevensJourney:
# Watched Queries
A bug existed where upstream changes would not reflect in watched queries. 

The table change update notifications were debounced, but the Kotlin flow `debounce` discards events which are triggered inside the debounce interval. This causes watched queries to be missed if other table changes occur (before) in the same interval - such as changes to `ps_buckets`. 

image

This PR accumulates table changes and emits the accumulated list of changed tables on the debounced interval.

https://github.com/powersync-ja/powersync-kotlin/assets/51082125/b0e82098-de5d-42d7-9554-a6c57ba98adf

# Schema indexes

The `powersync_replace_schema` call would throw an exception if a column index was included in the AppSchema. 

image

This was due to the schema incorrectly serialising the IndexColumn's column name as `column` instead of `name`.

This PR fixes the serialized field `name` and adds a link to the column definition in the index column definition in order to add the `type` field which is required (an exception is thrown if the `type` field is not included).

An index is also added to the Hello world demo app.


dialect/bin/main/com/powersync/sqlite/PowerSyncDialect.kt
```diff
@@ -0,0 +1,25 @@
+package com.powersync.sqlite
```
stevensJourney, line 1: This was generated after running the Hello World demo app. Not sure why it's not ignored or present in the main branch.

rkistner: Should be `bin` folder perhaps be .gitignored?

stevensJourney: I think so. I added it now to the ignore list now.

rkistner (dismissed):

rkistner (approved):

PR #12, opened 2024-04-08
fix: make field truly optional

DominicGBauer:
## Description
I saw this issue appear now that we fixed allowing `write_checkpoint` to be nullable.
```
SyncStream::streamingSync Error: kotlinx.serialization.MissingFieldException: Field 'write_checkpoint' is required for type with serial name 'com.powersync.sync.StreamingSyncCheckpointDiff', but it was missing
```

## Work Done
* Made `write_checkpoint` = null so that it is truly optional as explained in https://stackoverflow.com/questions/64796913/kotlinx-serialization-missingfieldexception. 

Note we could also implement this system wide to avoid similar issues in future https://github.com/Kotlin/kotlinx.serialization/blob/v1.3.0-RC/docs/json.md#explicit-nulls


DominicGBauer:
@Manrich121 Steven asked if you could comment on this please :) w.r.t. to setting `explicitNull`

stevensJourney (approved):

PR #13, opened 2024-04-11
chore: automate publishing

DominicGBauer:


plugins/sonatype/src/main/kotlin/com/powersync/plugins/sonatype/PublishToCentralPortalTask.kt
```diff
@@ -32,7 +32,7 @@ abstract class PublishToCentralPortalTask : DefaultTask() {
 
     companion object {
         const val UPLOAD_ENDPOINT =
-            "https://central.sonatype.com/api/v1/publisher/upload?publishingType=USER_MANAGED"
+            "https://central.sonatype.com/api/v1/publisher/upload?publishingType=AUTOMATIC"
```
Manrich121, line 35: This is fine as-is, but it might be worth it to make this configurable and use an enum in the future.

Like how the original Gradle plugin, we based our implement off, does it (https://gitlab.com/thebugmc/sonatype-central-portal-publisher/-/blob/master/src/main/java/net/thebugmc/gradle/sonatypepublisher/PublishingType.java?ref_type=heads)

Manrich121 (approved):

PR #16, opened 2024-04-17
[Fix] Sync Status

stevensJourney:
# Description

This fixes an issue where the `currentStatus` attribute of the PowerSync client would not get updated. This also adds the ability to receive a state flow for status change updates.

# Testing

The connection status has been added to the Hello World demo app.

https://github.com/powersync-ja/powersync-kotlin/assets/51082125/9497395a-2abe-4396-a4a0-4d63bc9e98c6

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
@@ -8,6 +8,7 @@ import com.powersync.db.crud.CrudBatch
 import com.powersync.db.crud.CrudTransaction
 import com.powersync.sync.SyncStatus
 import com.powersync.sync.SyncStream
+import kotlinx.coroutines.flow.StateFlow
 
 /**
  * A PowerSync managed database.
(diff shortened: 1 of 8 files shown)


core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
```diff
@@ -25,6 +26,11 @@ interface PowerSyncDatabase : ReadQueries, WriteQueries {
      */
     val currentStatus: SyncStatus
 
+    /**
+     * @returns a StateFlow which can be used to process sync state updates
+     */
+    fun getStatusFlow(): StateFlow<SyncStatus>
```
Manrich121, line 32: A common pattern is to have the class, `SyncStatus` in this case, have a method `asFlow` to provide a flow

This is an example from[SQLDelight](https://github.com/cashapp/sqldelight/blob/master/extensions/coroutines-extensions/src/commonMain/kotlin/app/cash/sqldelight/coroutines/FlowExtensions.kt#L37) adding `asFlow` to the `Query` class

Manrich121 (dismissed):
👍

Manrich121 (dismissed):

core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
stevensJourney: I updated the implementation to now have the `SyncStatus` provide a flow which emits data updates each time the status has been changed.

Manrich121 (approved):

Manrich121 (approved):

PR #19, opened 2024-05-06
Add disconnect and clear

mugikhan:
## Work Done
- Add disconnect and clear method

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -257,7 +257,6 @@ internal class PowerSyncDatabaseImpl(
 
     override suspend fun disconnect() {
         if(syncJob != null && uploadJob != null && syncJob!!.isActive && uploadJob!!.isActive) {
-            logger.debug { "Disconnecting" }
             //Wait for job to finish and then cancel
             syncJob?.cancelAndJoin()
             uploadJob?.cancelAndJoin()
--- a/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -260,7 +260,7 @@ internal class PowerSyncDatabaseImpl(
 
     override suspend fun disconnect() {
         if (syncJob != null && uploadJob != null && syncJob!!.isActive && uploadJob!!.isActive) {
-            //Wait for job to finish and then cancel with a CancellationException
+            //Wait for jobs to finish and then cancel with a CancellationException
             syncJob?.cancelAndJoin()
             uploadJob?.cancelAndJoin()
             syncStream = null
--- a/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -272,6 +272,7 @@ internal class PowerSyncDatabaseImpl(
             execute("DELETE FROM ${PsInternalTable.OPLOG}")
             execute("DELETE FROM ${PsInternalTable.CRUD}")
             execute("DELETE FROM ${PsInternalTable.BUCKETS}")
+            execute("DELETE FROM ${PsInternalTable.UNTYPED}")
 
             val tableGlob = if (clearLocal) "ps_data_*" else "ps_data__*"
             val existingTableRows = getAll(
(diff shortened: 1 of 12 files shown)


core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -247,4 +254,37 @@ internal class PowerSyncDatabaseImpl(
...
+    }
+
+    override suspend fun disconnectAndClear(clearLocal: Boolean) {
+        disconnect()
+
+        this.writeTransaction {
+            execute("DELETE FROM ${PsInternalTable.OPLOG}")
+            execute("DELETE FROM ${PsInternalTable.CRUD}")
+            execute("DELETE FROM ${PsInternalTable.BUCKETS}")
+
```
rkistner, line 275: Also clear `ps_untyped` (just picked up the issue in the JS and Dart SDKs)

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -247,4 +254,38 @@ internal class PowerSyncDatabaseImpl(
...
+        disconnect()
+
+        this.writeTransaction {
+            execute("DELETE FROM ${PsInternalTable.OPLOG}")
+            execute("DELETE FROM ${PsInternalTable.CRUD}")
+            execute("DELETE FROM ${PsInternalTable.BUCKETS}")
+            execute("DELETE FROM ${PsInternalTable.UNTYPED}")
+
+            val tableGlob = if (clearLocal) "ps_data_*" else "ps_data__*"
+            val existingTableRows = getAll(
```
Manrich121, line 278: There is a very similar query executed here https://github.com/powersync-ja/powersync-kotlin/blob/main/core/src/commonMain/kotlin/com/powersync/bucket/BucketStorage.kt#L47

This query should be added to the `PsInternalDatabase` and called from there.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -247,4 +254,38 @@ internal class PowerSyncDatabaseImpl(
...
+    override suspend fun disconnectAndClear(clearLocal: Boolean) {
+        disconnect()
+
+        this.writeTransaction {
+            execute("DELETE FROM ${PsInternalTable.OPLOG}")
+            execute("DELETE FROM ${PsInternalTable.CRUD}")
+            execute("DELETE FROM ${PsInternalTable.BUCKETS}")
+            execute("DELETE FROM ${PsInternalTable.UNTYPED}")
+
+            val tableGlob = if (clearLocal) "ps_data_*" else "ps_data__*"
```
Manrich121, line 277: Here is an `enum` and it should be updated to list both local and all data

https://github.com/powersync-ja/powersync-kotlin/blob/main/core/src/commonMain/kotlin/com/powersync/db/internal/PsInternalDatabase.kt#L36

Nit: Rather use a ternary instead of an inline `if`

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -87,7 +94,7 @@ internal class PowerSyncDatabaseImpl(
                 loggerFactory = loggerFactory
             )
 
-        scope.launch {
+        syncJob = scope.launch {
```
stevensJourney, line 97: We probably also need to call `disconnect` here before launching a new job. This is for cases where `connect` might be called multiple times. We want to ensure only 1 active connection is open at any point in time.

core/src/commonMain/kotlin/com/powersync/utils/Strings.kt
```diff
@@ -0,0 +1,7 @@
+package com.powersync.utils
+
+internal object Strings {
+    fun quoteIdentifier(s: String): String {
```
Manrich121, line 4: This can quote any string, not just identifiers, please rename this to `wrapInQuotes` or something similar. 

Also, instead of `internal object` consider extending `String`, like 

```
fun String.wrapInQuotes(): String {
    return if (this.startsWith("\"") && this.endsWith("\"")) {
        this
    } else {
        "\"$this\""
    }
}
```

Allowing it to be used like this:

```
var a = "foo"
var b = a.wrapInQuotes()
println(b) // Outputs: "foo"
```

rkistner: The quoting rules used for this are specific to SQLite identifiers - text literals use single quotes.

When extending String, will it still only available inside the library?

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
mugikhan: I added the query to `PsInternalDatabase`, unsure if we want it to publicly accessible though.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
mugikhan: Is an `if` statement not an expression in kotlin. I don't think kotlin supports ternary operators.

stevensJourney (dismissed):
Looks good to me. Nice to have this functionality now.

stevensJourney (approved):

PR #21, opened 2024-05-08
Add configurable retryDelayMs parameter

mugikhan:
## Work done

- Add a configurable retryDelay parameter to specify the delay between retrying after failure.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -85,7 +85,7 @@ internal class PowerSyncDatabaseImpl(
     }
 
     @OptIn(FlowPreview::class)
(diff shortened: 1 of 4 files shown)


core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -85,7 +85,7 @@ internal class PowerSyncDatabaseImpl(
     }
 
     @OptIn(FlowPreview::class)
-    override suspend fun connect(connector: PowerSyncBackendConnector) {
+    override suspend fun connect(connector: PowerSyncBackendConnector, retryDelayMs: Long) {
```
stevensJourney, line 88: It looks like the Dart SDK has `crudThrottleTime` which is used as the upload delay and `retryDelay` which is used as the sync stream download delay. Maybe we should specify both of those here.

mugikhan: Added that as a parameter as well.

stevensJourney (approved):
LGTM

PR #28, opened 2024-06-25
feat: show kdoc in swift and remove SKIE from PowerSyncBackendConnector functions

DominicGBauer:
## Description
* Show KDoc in Swift
* Remove SKIE from PowerSyncBackendConnector functions `uploadData` and `fetchCredentials` so that you are not required to add `__` before the functions as SKIE does not make changes to these functions. (This will require a change in Swift demo repo which will be added after this PR)

## Work Done
* Added flag which includes kdocs to Swift for each relevant module i.e PowerSync, core and Supabase connector. Note that the kdoc flag is also required in the `PowerSync` module otherwise it does not seem to compile the kdocs into Swift xcframework. 
* Added SKIE annotation to stop SKIE wrapping overwritten suspend functions `uploadData` and `fetchCredentials`. This required the inclusion of `configurationAnnotations = "0.8.2"`.

## Screenshots
**Example of KDoc in Swift with Xcode**
image

**Shows that __ is no longer necessary in Swift Demo**
image


core/build.gradle.kts
```diff
@@ -88,6 +89,12 @@ kotlin {
             languageSettings {
                 optIn("kotlinx.cinterop.ExperimentalForeignApi")
             }
+
+            val commonMain by sourceSets.getting {
+                dependencies {
+                    implementation(libs.configuration.annotations)
```
mugikhan, line 95: For my understanding, why can't this dependency be included in the `commonMain.dependencies`?

core/src/commonMain/kotlin/com/powersync/connectors/PowerSyncBackendConnector.kt
```diff
@@ -69,6 +70,7 @@ public abstract class PowerSyncBackendConnector {
      *
      * This token is kept for the duration of a sync connection.
      */
+    @SuspendInterop.Disabled
```
mugikhan, line 73: Why do we want to disable the suspend functionality of these functions?

core/build.gradle.kts
DominicGBauer: This was the recommendation from the docs on how to include it so I followed that. Do you think there's a better way to do this?

core/src/commonMain/kotlin/com/powersync/connectors/PowerSyncBackendConnector.kt
DominicGBauer: Here is what we received from the audit report:

`At the moment, SKIE’s suspend bridging introduces a prefix of two underscores to the original
suspend method. That means that users have to implement __fetchCredentials() and
__uploadData(database:) in their code. Since these methods aren’t supposed to be called
from Swift, we recommend turning SKIE off for the PowerSyncBackendConnector class.
SKIE currently can’t bridge the implementation back to Kotlin, so it doesn’t bring the
cancellation/cooperative concurrency benefit. You can disable SKIE’s suspend feature for this class
through either annotations or Gradle configuration in the powersyncswift module.`

core/build.gradle.kts
mugikhan: In the same file a couple of lines down `commonMain.dependencies` there are dependencies declared there. Can't we also include this there?

DominicGBauer: Done, thanks!

stevensJourney (approved):

PR #29, opened 2024-06-26
chore: replace PowerSyncDatabaseBuilder with global factory

DominicGBauer:
## Description
Simplify API by removing the `PowerSyncDatabaseBuilder` and replacing it with a global factory function `PowerSyncDatabase` so that we give developers a better more intuitive  experience.

## Work Done
* Removed `PowerSyncDatabaseBuilder`
* Added `PowerSyncDatabaseFactory.kt` which houses the global factory function `PowerSyncDatabase`
* Updated demos to use new global factory function

# Testing
**Swift**

https://github.com/powersync-ja/powersync-kotlin/assets/46312751/260baf43-39ca-4656-8cf9-c0e22701eb9a

**Kotlin Android**
 
https://github.com/powersync-ja/powersync-kotlin/assets/46312751/0f9037b1-817e-4b8d-bc21-1906cba5115f


core/src/commonMain/kotlin/com/powersync/PowerSyncDatabaseFactory.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+@OptIn(DelicateCoroutinesApi::class)
+@DefaultArgumentInterop.Enabled
+public fun createPowerSyncDatabase(
+    factory: DatabaseDriverFactory,
+    schema: Schema,
+    dbFilename: String = DEFAULT_DB_FILENAME,
+    scope: CoroutineScope = GlobalScope
+): PowerSyncDatabase {
+    return PowerSyncDatabaseImpl(
```
stevensJourney, line 23: For context, why is the `createPowerSyncDatabase` required? Could user's not just construct the client as on line 23?

DominicGBauer: Are you asking if we want to users to use `PowerSyncDatabaseImpl` to create the DB? I guess they could but I think that could be an anti-pattern in Kotlin. I'm not sure you're meant to expose implementations to the public. Based on the feedback in the audit this appears to be the way to do it.

DominicGBauer: Also note I've renamed `createPowerSyncDatabase` to `PowerSyncDatabase` as this seems to be the way to do it in Kotlin w.r.t a factory generating an instance of a class

stevensJourney: I like the latest changes

stevensJourney (approved):

PR #30, opened 2024-07-02
chore: improve logging

DominicGBauer:
## Description
Improve the logging experience for users. 

Currently, all the logs are exposed which may result in sensitive information being logged resulting in a security vulnerability. To improve this I have changed the current logger Canard to use touchlab's [Kermit](https://kermit.touchlab.co/docs/) logger and if a user does not provide a logger I have set some sensible log levels to be displayed when it is `debug` or `release` build. The reason for this is 

1. It allows the user to set the global logger settings
2. It is more popular which increases the likelihood of users already using it
3. The experience is now more aligned with the other SDK's
4. It is extendable by the user but has sensible defaults set if they opt not to

## Work Done
* Added a build config flag which is used to determine if it is a debug build or not
* Swapped out current logger for new Kermit logger
* Set sensible log defaults if user does not provide a logger

## Testing
I have tested this on both iOS and Android in `debug` and `release` builds and it works as expected i.e. all logs are shown in `degub` and only warnings and above are shown in `release`

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -102,8 +102,8 @@ kotlin {
             implementation(libs.kotlinx.datetime)
             implementation(libs.stately.concurrency)
             implementation(libs.bundles.sqldelight)
-            implementation(libs.canard)
             implementation(libs.configuration.annotations)
+            implementation(libs.kermit)
         }
 
         androidMain.dependencies {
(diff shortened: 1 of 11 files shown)


stevensJourney (approved):
Looks good to me!

PR #31, opened 2024-07-04
chore: remove global scope

DominicGBauer:
## Description
There is an unnecessary use of global scope in `prefetchCredentials`.

## Work Done
Removed `GlobalScope` and replaced it with a job that works on both iOS and Android.


core/src/commonMain/kotlin/com/powersync/connectors/PowerSyncBackendConnector.kt
```diff
@@ -47,16 +43,14 @@ public abstract class PowerSyncBackendConnector {
...
      * This may be called before the current credentials have expired.
      */
-    @OptIn(DelicateCoroutinesApi::class)
     public suspend fun prefetchCredentials(): PowerSyncCredentials? {
-        fetchRequest = fetchRequest ?: GlobalScope.async {
-            fetchCredentials().also { value ->
-                cachedCredentials = value
-                fetchRequest = null
-            }
+        fetchRequest = fetchRequest ?: fetchCredentials().also { value ->
```
stevensJourney, line 47: I'm not following the logic here anymore. Previously `fetchRequest` was a `Defered<PowerSyncCredentials>` which in my understanding was used to keep track of a pending operation of fetching credentials. I thought this was to ensure that the `fetchCredentials` function is not called multiple times simultaneously. 

But in this new implementation it seems like `fetchRequest` is the result value of `PowerSyncCredentials` which is cleared after `fetchCredentials` resolves - but this value is also returned at the end of this function. 

Can you add some comments to explain what this is doing?

stevensJourney (approved):

PR #33, opened 2024-07-10
chore: turn SKIE off for suspend functions

DominicGBauer:
## Description
When SKIE bridges the code from Kotlin to Swift it makes a few changes to suspend functions that have resulted in quite a few issues. 

The crux of the issue is that `async` functions in Swift are running on the main thread while `suspend` functions that are transformed using SKIE to Swift are run on a background thread. This results in transactions starting on the main thread while the database execute functions in the transaction are run on a background thread leading to deadlocks and changes being rejected as indicated by the audit report. Further this seems to have also had implications on token refreshes and once removed has fixed those issues.

This PR turns of SKIE changes to Kotlin `suspend` functions. From some manual testing on the demo app there does not appear to be any issue or slowness as a result of this change. This will require a change PR in the Swift demo once this is merged an released as async functions won't need a `__` anymore.

### DIFF
--- a/PowerSync/build.gradle.kts
+++ b/PowerSync/build.gradle.kts
@@ -1,4 +1,5 @@
 import co.touchlab.faktory.versionmanager.TimestampVersionManager
+import co.touchlab.skie.configuration.SuspendInterop
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 
 plugins {
@@ -31,6 +32,14 @@ kotlin {
     sourceSets {
         commonMain.dependencies {
             api(project(":core"))
+        }
+    }
+}
+
+skie {
+    features {
+        group {
+            SuspendInterop.Enabled(false) // or false
         }
     }
 }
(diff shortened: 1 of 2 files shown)


PowerSync/build.gradle.kts
```diff
@@ -35,6 +36,14 @@ kotlin {
     }
 }
 
+skie {
+    features {
+        group {
+            SuspendInterop.Enabled(false) // or false
```
mugikhan, line 42: Nitpick: The comment doesn't make sense here, maybe it can explain why we are disabling suspend for SKIE.

DominicGBauer: Sorry I missed that. I've removed the comment and replaced it with an explanation.

mugikhan (approved):

PowerSync/build.gradle.kts
mugikhan: Another little note: I think `compiling` is a better word to use here rather than `implementing`.

DominicGBauer: Oh shucks sorry missed this before I merged. Will include it in the next change I make

DominicGBauer: Oh no wait there isn't an issue at compile time. The issue happens when you implement parts of the SDK in Swift.

PR #38, opened 2024-07-25
WIP: JVM support

henwoods:
Currently non-functional and only tested on OSX. I've not much experience bundling native libs so probably there is convention errors and hackiness that needs to be cleaned up with the build/make/gradle configs and the library loader.

Notes:
 - Currently crashes on start up with `[SQLITE_ERROR] SQL error or missing database (no such function: powersync_rs_version)`, guessing this is because we need a way to support extensions for the JVM driver. Is this handled by requery on Android?
 - Library loader is (hackily?) just pulling the shared library out to a temporary file, worth exploring other options.
  
 TODOs:
-  [ ] Didn't implement a way to determine if debug/release build yet
 - [ ] `sqlight_bindings.cpp` is copy/pasta from Android, would be good to pull it up so it can be shared 
 - [ ] No logic for storing the database in a shared user folder yet, when run from Android Studio the database is just dumped in the build folder

Usage:
 - Noted above, Windows/Linux will require at least implementing a library loader with the appropriate shared library extension.
 - Gradle Sync 
 - Install locally:
   - Disable release signing (`RELEASE_SIGNING_ENABLED=false`)
   - Run `publishToMavenLocal` task 
 - Open `powersync-hello` and create a Gradle run configuration
    - Select the `hello-powersync:desktop` module 
    - Set the Run field to `run -DmainClass=com.powersync.demo.MainKt`
 - Run the new config

Expect the app to crash with the missing `powersync_rs_version` function error

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -1,6 +1,7 @@
-import de.undercouch.gradle.tasks.download.Download
-import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import com.powersync.plugins.sonatype.setupGithubRepository
+import de.undercouch.gradle.tasks.download.Download
+import org.apache.tools.ant.taskdefs.condition.Os
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 
 plugins {
     alias(libs.plugins.kotlinMultiplatform)
@@ -14,8 +15,15 @@ plugins {
 val sqliteVersion = "3450000"
 val sqliteReleaseYear = "2024"
 
-val sqliteSrcFolder =
-    project.layout.buildDirectory.dir("interop/sqlite").get()
+val osName = when {
+    Os.isFamily(Os.FAMILY_WINDOWS) -> "Windows"
+    Os.isFamily(Os.FAMILY_MAC) -> "macOS"
+    Os.isFamily(Os.FAMILY_UNIX) -> "Linux"
+    else -> "Unknown"
+}
+
+val sqliteSrcFolder = project.layout.buildDirectory.dir("interop/sqlite").get()
+val jvmNativeBuildFolder = "${layout.buildDirectory.get().asFile}/native/powersync-sqlite"
 
 val downloadSQLiteSources by tasks.registering(Download::class) {
     val zipFileName = "sqlite-amalgamation-$sqliteVersion.zip"
@@ -59,6 +67,19 @@ val buildCInteropDef by tasks.registering {
     outputs.files(defFile)
 }
 
+tasks.create<Exec>("buildNative") {
+    group = "build"
+
+    outputs.dir(jvmNativeBuildFolder)
+    workingDir = file(jvmNativeBuildFolder)
+
+    environment("TARGET", "$osName/cmake")
+    environment("SOURCE_PATH", "$projectDir/src/jvmMain/cpp")
+    environment("INTEROP_PATH", "$sqliteSrcFolder")
+
+    commandLine("$projectDir/src/jvmMain/cpp/build.sh")
+}
+
 kotlin {
     androidTarget {
         publishLibraryVariants("release", "debug")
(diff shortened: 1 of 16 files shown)


core/build.gradle.kts
```diff
@@ -59,6 +67,19 @@ val buildCInteropDef by tasks.registering {
     outputs.files(defFile)
 }
 
+tasks.create<Exec>("buildNative") {
```
henwoods, line 70: Run `build.sh` to build the binary and save it to `build/native/powersync-sqlite/${OS}/output`

core/build.gradle.kts
```diff
@@ -67,6 +88,13 @@ kotlin {
     iosX64()
     iosArm64()
     iosSimulatorArm64()
+    jvm {
+        val processResources = compilations["main"].processResourcesTaskName
```
henwoods, line 92: Bundle the binary into the jar

core/src/jvmMain/cpp/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
henwoods, line 1: Not sure whats the convention for the location of build files 🤔

core/src/jvmMain/cpp/sqlite_bindings.cpp
```diff
@@ -0,0 +1,91 @@
+#include <jni.h>
```
henwoods, line 1: Mentioned in the PR note, this is copied from `androidMain`, would be good to share the file

core/src/jvmMain/kotlin/BuidConfig.kt
```diff
@@ -0,0 +1,6 @@
+public actual object BuildConfig {
+    public actual val isDebug: Boolean
+        // TODO: need to determine a good way to set this on JVM presuming we don't want to bundle BuildKonfig in the
+        //  library.
+        get() = true
```
henwoods, line 5: Didn't want to add the dependency on `BuildKonfig` to the library, will explore if there is an alternative.

core/src/jvmMain/kotlin/com/powersync/DatabaseDriverFactory.jvm.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public actual fun createDriver(
+        scope: CoroutineScope,
+        dbFilename: String
+    ): PsSqlDriver {
+        val schema = InternalSchema.synchronous()
+        this.driver = PsSqlDriver(
+            scope = scope,
+            driver = JdbcSqliteDriver("jdbc:sqlite:$dbFilename", Properties(), schema)
```
henwoods, line 37: I think this is the cause of the crash, not loading extensions so the `powersync_rs_version` isn't available

core/src/jvmMain/kotlin/com/powersync/LibraryLoader.kt
```diff
@@ -0,0 +1,38 @@
+package com.powersync
+
+import com.powersync.LibraryLoader.Companion.SQLITE_BINARY_FILENAME
+import java.io.File
+import java.io.InputStream
+
+private interface LibraryLoader {
```
henwoods, line 7: AI recommended this approach so I'm sure there is a better way 😂

DominicGBauer:
Thanks @henwoods for getting this going, I have gone through the PR and run into the same issue as you've pointed out. 

I think what might help is that we already release powersync-sqlite-core https://github.com/powersync-ja/powersync-sqlite-core/releases for different targets i.e. linux, windows etc which will be useful for here.

Here's an example of how it has been implemented in Flutter [https://github.com/powersync-ja/powersync.dart/blob/54f7a69d109d22358f3b8e3fda285d[…]2e0fb94f60/packages/powersync_flutter_libs/linux/CMakeLists.txt](https://github.com/powersync-ja/powersync.dart/blob/54f7a69d109d22358f3b8e3fda285d2e0fb94f60/packages/powersync_flutter_libs/linux/CMakeLists.txt#L28). 

https://github.com/powersync-ja/powersync.dart/blob/54f7a69d109d22358f3b8e3fda285d2e0fb94f60/packages/powersync/lib/src/open_factory/native/native_open_factory.dart#L25

`enableExtension()` seems to be the same as `System.loadLibrary(...)`.

I will have more time to look at this again next week.

Shabinder:
is this currently being worked on ?
would be great to have this, and sqldelight for using it with ease.

DominicGBauer:
@Shabinder we have noted your interest but we are not currently looking at this. We are open to anyone in the community continuing with the PR to get it complete.

kobiebotha:
Closing this as it has been superseded by https://github.com/powersync-ja/powersync-kotlin/pull/88

PR #45, opened 2024-08-20
chore: use callback in transactions

DominicGBauer:
## Description
Adopt callback pattern for `readTransaction` and `writeTransaction`. 

## Work Done
Changed `readTransaction` and `writeTransaction` functionality to use a callback such that users can now use for example
db.writeTransaction { tx ->
  tx.execute()
  tx.get()
  ...
}

## Video
https://github.com/user-attachments/assets/75720202-95f3-45eb-b939-cce6bde63cff


stevensJourney (approved):
This looks like a good start towards having better defined multiple connections and locks in the future. I'm happy with this from my side.

PR #57, opened 2024-09-20
feat: add user agent

DominicGBauer:
## Description
Add `User-Agent` header to requests to identify the sdk being used. 

**Note:** Currently this will not differentiate based on the Kotlin or Swift SDK and will only show Kotlin SDK in the user agent. This is because we need to:
1. Update publish plugin to publish individual modules 
2. Create a separate build and publish github workflow for the Swift SDK  and add `-Pbuildkonfig.flavor=swift-sdk` as a flag.
3. This flag will then change the Library to reflect `powersync-swift` instead of `powersync-kotlin`

## Work Done
* Added the `User-Agent` header to requests
* Added BuildKonfig module which will be used to differentiate between Kotlin and Swift sdks.
* Removed `User-Id` header

## Testing
image

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
+++ b/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
@@ -8,6 +8,8 @@ import com.powersync.bucket.Checkpoint
 import com.powersync.bucket.WriteCheckpointResponse
 import co.touchlab.stately.concurrency.AtomicBoolean
 import com.powersync.connectors.PowerSyncBackendConnector
+import com.powersync.core.BuildKonfig.LIBRARY_VERSION
+import com.powersync.core.BuildKonfig.LIBRARY_NAME
 import com.powersync.utils.JsonUtil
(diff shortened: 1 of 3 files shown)


core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -174,6 +180,7 @@ internal class SyncStream(
             headers {
                 append(HttpHeaders.Authorization, "Token ${credentials.token}")
                 append("User-Id", credentials.userId ?: "")
```
rkistner, line 182: While you're working on headers, could you also remove the "User-Id" one in both places? It is never used by the service, and not included by other SDKs.

DominicGBauer: Will do

simolus3:
A user agent has been added to requests by default in https://github.com/powersync-ja/powersync-kotlin/pull/192.

PR #60, opened 2024-09-23
feat(core): add client sync rules paramters

DominicGBauer:
## Description
Add `params` option in `connect` method of `PowerSyncDatabase` to allow users to set sync rules parameters from the client. To ensure type safety a new sealed class `JsonParam` was created so that `params` accepts a `Map<String, JsonParam?>?` which will prevent users erroneously including invalid parameters. Example usage:

```kotlin
    val params = JsonParam.Map(
          mapOf(
              "name" to JsonParam.String("John Doe"),
              "age" to JsonParam.Number(30),
              "isStudent" to JsonParam.Boolean(false)
         )
     )

   connect(
         connector = connector,
         crudThrottleMs = 2000L,
         retryDelayMs = 10000L,
         params = params
     )
```

## Testing
I set new sync rules 
```
bucket_definitions:
  list_buckets:
    accept_potentially_dangerous_queries: true
    parameters: select request.parameters() ->> 'listId' as list_id
    data:
      - SELECT * FROM lists where id = bucket.list_id
```
and passed a `listId` from the client side 
```
db.connect(supabase, params = mapOf("listId" to "c3a57c9a-06e0-490a-9003-66cd838f7bde".toJsonParam()))
```

https://github.com/user-attachments/assets/38c52a9f-7be7-46c0-90f0-bb23de9f5162


core/src/commonMain/kotlin/com/powersync/utils/Json.kt
```diff
@@ -12,3 +15,18 @@ internal object JsonUtil {
...
+internal fun convertMapToJson(map: Map<String, Any>?): JsonObject {
+    if (map == null) return JsonObject(emptyMap())
+
+    val result = map.mapValues { (_, value) ->
+        when (value) {
+            is Int -> JsonPrimitive(value)
+            is Long -> JsonPrimitive(value)
+            is Double -> JsonPrimitive(value)
+            is Boolean -> JsonPrimitive(value)
+            is String -> JsonPrimitive(value)
```
stevensJourney, line 27: Are arrays/lists supported?

DominicGBauer: Yeah you're right I've added support for Arrays, Lists and Maps now as well

core/src/commonMain/kotlin/com/powersync/utils/Json.kt
```diff
@@ -12,3 +17,32 @@ internal object JsonUtil {
     }
 }
 
+internal fun convertMapToJson(map: Map<String, Any?>?): JsonObject {
+    if (map == null) return JsonObject(emptyMap())
+
+    val result = map.mapValues { (_, value) ->
```
mugikhan, line 23: Is there custom mapping necessary? Would `JsonObject(map)` not convert it to a `JsonObject` with the correct types? Does it fail to convert some types to string in the `JsonObject`?

Chriztiaan (approved):

core/src/commonMain/kotlin/com/powersync/utils/Json.kt
DominicGBauer: `JsonObject` requires a `Map<String, JsonElement>` while we are receiving a `Map<String, Any?>` so we need to convert the values to a JsonElement before we can run `JsonObject` on the map, unless I am missing something?

stevensJourney (approved):

PR #65, opened 2024-10-07
Feat/add wait for first sync

DominicGBauer:
## Description
Add public `waitForFirstSync` function that will only resolve after the first sync

## Work Done
* Added `waitForFirstSync` function to `PowerSyncDatabaseImpl` and `PowerSyncDatabase` interface
* Fixed issue in SyncStatus output where a colon should have been an equals sign

## Testing
Added `waitForFirstSync` before and after `updateHasSynced` function. 
1. If it is placed before then it stops the app as hasSynced is never updated. 
2. If its placed afterwards then the app continues as expected.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
@@ -23,6 +23,11 @@ public interface PowerSyncDatabase : Queries {
     public val currentStatus: SyncStatus
 
     /**
+     * Suspend function that resolves when the first sync has occurred
+     */
+    public suspend fun waitForFirstSync()
+
+    /**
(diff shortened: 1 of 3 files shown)


stevensJourney:
Just checking here:
"If it is placed before then it stops the app as hasSynced is never updated."
It might be confusing me since `updateHasSynced` is an internal function. The testing methodology here might be modifying the SDK behaviour. 

In normal conditions, does the `hasSynced` eventually update to true when the `lastSyncedAt` updates? This should eventually cause `waitForFirstSync` to resolve.

DominicGBauer:
Yes `waitForFirstSync` resolves after `hasSynced` has been updated.

mugikhan (approved):
LGTM

PR #73, opened 2024-10-29
fix: custom write checkpoint logic

stevensJourney:
# Overview

This aligns the logic for Custom Write Checkpoints with the JS and Dart SDKs by correcting an error in the logical tests.

For reference see:
- Dart [implementation](https://github.com/powersync-ja/powersync.dart/blob/fb2f312247d44f74efaaa097c090190f9514e607/packages/powersync/lib/src/database/powersync_db_mixin.dart#L406)
- JS [implementation](https://github.com/powersync-ja/powersync-js/blob/30825f3e6137a86cf37a2e393e9e193bcf1debc4/packages/common/src/client/AbstractPowerSyncDatabase.ts#L577)

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -252,7 +252,7 @@ internal class PowerSyncDatabaseImpl(
         writeTransaction { tx ->
             internalDb.queries.deleteEntriesWithIdLessThan(lastTransactionId.toLong())
 
-            if (writeCheckpoint != null && bucketStorage.hasCrud()) {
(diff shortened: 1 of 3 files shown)


core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -252,7 +252,7 @@ internal class PowerSyncDatabaseImpl(
         writeTransaction { tx ->
             internalDb.queries.deleteEntriesWithIdLessThan(lastTransactionId.toLong())
 
-            if (writeCheckpoint != null && bucketStorage.hasCrud()) {
+            if (writeCheckpoint != null && !bucketStorage.hasCrud()) {
```
DominicGBauer, line 255: Can we add a test to check this?

PR #81, opened 2024-12-03
fix: transactions resulting in error when using different threads

DominicGBauer:
## Description
This PR removes disabling SKIE suspendInterop for all functions. It moves the transaction object out of the generator function as using this function was the root cause of why the thread was changing when a transaction was being run in the Swift SDK.

## Testing
Refer to https://github.com/powersync-ja/powersync-swift/pull/3 where tests were created to test database transactions.


stevensJourney (approved):
This work-around is probably fine for now, but we might run in to issues later when we require using different SQLite connections for concurrent transactions.

PR #83, opened 2024-12-09
chore: change swift package name

DominicGBauer:
## Description
As we are moving to a new native swift package we want that package to use the name `PowerSync`. This PR changes the kotlin generated SDK's name from `PowerSync` to `PowerSyncKotlin` so that the new native Swift can use the name otherwise it results in a collision.

A PR will then be created in the new swift repo to use the PowerSyncKotlin package and update the package details based on this PR.


stevensJourney (approved):
Happy from my side.

PR #88, opened 2024-12-16
JVM support

SalomonBrys:
Support for the JVM on the following platforms:
- Linux Aarch64
- Linux X64
- MacOS Aarch 64
- MacOS X64
- Windows X64

Here are some highlights of this PR:
- Gradle downloads powersync core extension binaries [directly from its release](https://github.com/powersync-ja/powersync-sqlite-core/releases).
- Supports 2 build mode:
  - By default, only builds for the current platform.
  - With `-Ppowersync.binaries.provided="true"`, **do not build the native binaries** but expect all 5 of them in the `core/binaries/desktop` directory.

The CI builds each binary on its own platform and then collects them for the "final" build with is built with the `powersync.binaries.provided` property (and therefore uploads a single JVM dependency that contains binaries for all 5 platforms).

- The `PSJdbcSqliteDriver.enableWriteAheadLogging` function does what it says, but is not used as enabling WAL made the app crash with a `SQLITE_BUSY` error. Enabling this optimisation is outside the scope of this PR.


DominicGBauer:
@SalomonBrys I am getting the following errors when trying to run the hello-powersync using desktop run.

```
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':powersync-kotlin:core:cmakeJvmConfigureAarch64'.
> A problem occurred starting process 'command 'cmake''
```
image

What am I missing?

SalomonBrys:
Is CMake installed on your system ?
Is it in your system's PATH ?

rkistner:
Is there a way to avoid cross-compiling by default?

On my Linux machine I do have cmake installed, but I'm getting this error when running the demo:

```
> Task :powersync-kotlin:core:cmakeJvmConfigureAarch64 FAILED
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:3 (project):
  The CMAKE_CXX_COMPILER:

    aarch64-linux-gnu-g++

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!

> Task :powersync-kotlin:core:downloadPowersyncDesktopBinaries UP-TO-DATE
Download https://github.com/powersync-ja/powersync-sqlite-core/releases/download/v0.3.6/libpowersync_aarch64.so
Download https://github.com/powersync-ja/powersync-sqlite-core/releases/download/v0.3.6/libpowersync_x64.so

```

Commenting out this line resolves the issue:

```kt
registerCMakeTasks("aarch64", "CMAKE_C_COMPILER=aarch64-linux-gnu-gcc", "CMAKE_CXX_COMPILER=aarch64-linux-gnu-g++"),
```

Unless I'm doing a full build, I'd prefer if only my current architecture was used. I'm not sure whether the same issue affects MacOS, or only Linux.

Apart from this, the demo app runs fine. I haven't thoroughly reviewed the code yet, but we should be able to fix any other issues over time.

SalomonBrys:
As requested, I have modified the Gradle build script so that, by default, only the current OS and architecture are built.
I've updated the description of this PR to explain all 3 build modes.

rkistner (approved):

DominicGBauer (approved):

PR #91, opened 2024-12-19
fix: race condition in transaction

DominicGBauer:
## Problem
We were sporadically encountering `IllegalStateException` with the message "Transaction objects must be used only within the transaction lambda scope." This typically indicates a race condition where transaction objects are being accessed across different threads or coroutine contexts. I initially thought it was because of a transaction in BucketStorageImpl was using the wrong context (which was addressed in this PR https://github.com/powersync-ja/powersync-kotlin/pull/89) but this was not the case as the issue persisted.

## Root Cause
Database operations were not consistently running on the same dispatcher, which could lead to transaction objects being accessed from different threads. SQLDelight requires transaction operations to maintain thread confinement for safety.

## Solution
Added a dedicated IO dispatcher for all database operations by modifying the database driver creation:
```kotlin
val dbDispatcher = Dispatchers.IO
val dbScope = CoroutineScope(scope.coroutineContext + dbDispatcher)
```
This will also improve DB operation performance as the operations are now always running on the correct dispatcher.

## Testing
Testing in the Swift demo no longer produces this error (although it's tough to say that this is definitively resolved since the error was sporadic).

### DIFF
--- a/core/src/iosMain/kotlin/com/powersync/DatabaseDriverFactory.ios.kt
+++ b/core/src/iosMain/kotlin/com/powersync/DatabaseDriverFactory.ios.kt
@@ -17,6 +17,8 @@ import kotlinx.cinterop.asStableRef
 import kotlinx.cinterop.staticCFunction
 import kotlinx.cinterop.toKString
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.IO
 
 @Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
 @OptIn(ExperimentalForeignApi::class)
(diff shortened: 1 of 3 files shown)


core/src/iosMain/kotlin/com/powersync/DatabaseDriverFactory.ios.kt
```diff
@@ -51,10 +53,13 @@ public actual class DatabaseDriverFactory {
         scope: CoroutineScope,
         dbFilename: String,
     ): PsSqlDriver {
+        val dbDispatcher = Dispatchers.IO
+        val dbScope = CoroutineScope(scope.coroutineContext + dbDispatcher)
```
mugikhan, line 57: I'd be careful with using the `CoroutineScope` interface here. The docs mention `Manual implementation of this interface is not recommended, implementation by delegation should be preferred instead.` https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/

It might be better to use `coroutineScope`.
Some info from my GPT friend: `coroutineScope` is used for Structured concurrency. Structured concurrency guarantees:

- When a scope cancels, all of its coroutines cancel.
- When a suspend function returns, all of its work is done.
- When a coroutine errors, its caller or scope is notified.
The coroutineScope unlike the CoroutineScope, will suspend itself until all coroutines started inside of it are complete. Indeed it creates a local scope and guarantees that all child coroutines will be completed before the return instruction.
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html

mugikhan (dismissed):

mugikhan (approved):

Issue #102, opened 2025-01-12
`invalidateCredentials` of `PowerSyncBackendConnector` should not be called on any streamSync error


benoitletondor:
See https://github.com/powersync-ja/powersync-kotlin/blob/main/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt#L103

The following code calls `invalidateCredentials` every time there's an error, no matter the error. If you have no network, the credentials are invalidated every X seconds, resulting in a lot of useless queries (that will fail since there's no network anyway).

Also, the code just above is probably wrong too, you want to rethrow the `CancellationException` if you catch it, not just avoiding logging in case of it. See https://medium.com/mobilepeople/how-to-break-a-coroutine-cancellation-12e968c31642

DominicGBauer:
Hi @benoitletondor thanks for raising this as well. I have created a PR that should improve this https://github.com/powersync-ja/powersync-kotlin/pull/104.

PR #104, opened 2025-01-14
feat: do not invalidate credentials on io error

DominicGBauer:
## Description
This aims to resolve https://github.com/powersync-ja/powersync-kotlin/issues/102 where credentials are invalidated even though there is no network. It appears this was included before invalidation was properly handled and can be removed.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
+++ b/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
@@ -31,7 +31,6 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
 import kotlinx.datetime.Clock
-import kotlinx.io.IOException
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
@@ -102,11 +101,6 @@ internal class SyncStream(
                 }
 
                 logger.e(e) { "Error in streamingSync: $e" }
-                // If there is a network issue don't invalidate the credentials
-                if (e !is IOException) {
-                    invalidCredentials = true
-                }
-
                 status.update(
                     downloadError = e,
                 )
--- a/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
+++ b/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
@@ -97,11 +97,11 @@ internal class SyncStream(
 //                    break;
 //                }
             } catch (e: Exception) {
-                // If the coroutine was cancelled, don't log an error
-                if (e !is CancellationException) {
-                    logger.e(e) { "Error in streamingSync: $e" }
+                if (e is CancellationException) {
+                    throw e
                 }
 
+                logger.e(e) { "Error in streamingSync: $e" }
                 // If there is a network issue don't invalidate the credentials
                 if (e !is IOException) {
                     invalidCredentials = true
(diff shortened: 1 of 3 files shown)


core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -96,11 +97,16 @@ internal class SyncStream(
...
-                if (e !is CancellationException) {
-                    logger.e(e) { "Error in streamingSync" }
+                if (e is CancellationException) {
+                    throw e
                 }
-                invalidCredentials = true
+
+                logger.e(e) { "Error in streamingSync: $e" }
+                // If there is a network issue don't invalidate the credentials
+                if (e !is IOException) {
```
stevensJourney, line 106: Could we match the Dart implementation and invalidate if we [receive](https://github.com/powersync-ja/powersync.dart/blob/bd22580265eb1b690d19e629c421f3233c17adf1/packages/powersync_core/lib/src/streaming_sync.dart#L263) a 401 HTTP response code or if we receive a message that the token is expiring [soon](https://github.com/powersync-ja/powersync.dart/blob/bd22580265eb1b690d19e629c421f3233c17adf1/packages/powersync_core/lib/src/streaming_sync.dart#L408)?

DominicGBauer: The 401 invalidation is happening here https://github.com/powersync-ja/powersync-kotlin/blob/c74857ea9a155c71d070e978d3a2b4173daca6d6/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt#L178 and the token expiring invalidation is happening here https://github.com/powersync-ja/powersync-kotlin/blob/c74857ea9a155c71d070e978d3a2b4173daca6d6/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt#L393

stevensJourney: If we already invalidate in those places, what is the reason for invalidating in this section of the code?

DominicGBauer: Not sure, it seems to have been added in the beginning probably before the other invalidations were added. I will remove it.

core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -96,11 +96,11 @@ internal class SyncStream(
 //                    break;
 //                }
             } catch (e: Exception) {
-                // If the coroutine was cancelled, don't log an error
-                if (e !is CancellationException) {
-                    logger.e(e) { "Error in streamingSync" }
+                if (e is CancellationException) {
+                    throw e
```
stevensJourney, line 100: Whats the reason for rethrowing here? How would the error surface? What would the difference be compared to breaking out of the loop?

DominicGBauer: I'm following this guide https://medium.com/mobilepeople/how-to-break-a-coroutine-cancellation-12e968c31642 which explains that swallowing a cancellation exception may lead to unintended consequences as the coroutine will continue it's execution until the next suspend call. 

The error surfacing would depend on what resulted in the cancellation. 

The difference is that an error will be thrown with the reason for the error whereas breaking out of the loop will just break out of the loop without a reason which could lead to difficult to debug issues.

stevensJourney: Thanks for the context. Sounds good to me.

stevensJourney (approved):

PR #105, opened 2025-01-16
feat: add macos to xcframework

DominicGBauer:
## Description
This aims to include macOS to the swift SDK

- [ ] Add macOS to Package.swift
- [ ] Fix this issue "the path does not point to a valid framework: /Users/dominic/Desktop/powersync/powersync-kotlin/PowerSyncKotlin/build/PowerSyncKotlinXCFrameworkTemp/fatframework/debug/macos/PowerSyncKotlin.framework"


simolus3:
We've since added macOS to the Swift SDK with a similar approach.

PR #107, opened 2025-01-24
Synchronized transactions

kpgalligan:
* Remove async from sqldelight config
* Remove suspend from the transaction block and methods available on the PowerSyncTransaction interface
* Some access cleanup to ensure all direct db access is handled by `InternalDatabaseImpl`
* Move calls to the actual db off of caller thread inside `InternalDatabaseImpl`

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabase.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabase.kt
@@ -2,13 +2,11 @@ package com.powersync.db.internal
 
 import app.cash.sqldelight.db.Closeable
 import com.persistence.PowersyncQueries
-import com.powersync.PsSqlDriver
 import com.powersync.db.Queries
 import com.powersync.persistence.PsDatabase
 import kotlinx.coroutines.flow.Flow
 
 internal interface InternalDatabase : Queries, Closeable {
-    val driver: PsSqlDriver
     val transactor: PsDatabase
     val queries: PowersyncQueries
 
--- a/core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
@@ -2,8 +2,6 @@ package com.powersync.db.internal
 
 import app.cash.sqldelight.ExecutableQuery
 import app.cash.sqldelight.Query
-import app.cash.sqldelight.async.coroutines.awaitAsList
-import app.cash.sqldelight.async.coroutines.awaitAsOneOrNull
 import app.cash.sqldelight.coroutines.asFlow
 import app.cash.sqldelight.coroutines.mapToList
 import app.cash.sqldelight.db.QueryResult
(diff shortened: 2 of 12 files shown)


core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabase.kt
```diff
@@ -2,13 +2,11 @@ package com.powersync.db.internal
 
 import app.cash.sqldelight.db.Closeable
 import com.persistence.PowersyncQueries
-import com.powersync.PsSqlDriver
 import com.powersync.db.Queries
 import com.powersync.persistence.PsDatabase
 import kotlinx.coroutines.flow.Flow
 
 internal interface InternalDatabase : Queries, Closeable {
-    val driver: PsSqlDriver
```
kpgalligan, line 11: I removed this from the interface. Nobody was calling it, and really shouldn't be. Actual db access should be handled entirely by `InternalDatabaseImpl`.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
```diff
@@ -14,44 +12,50 @@ import com.powersync.PsSqlDriver
...
 internal class InternalDatabaseImpl(
-    override val driver: PsSqlDriver,
+    private val driver: PsSqlDriver,
     private val scope: CoroutineScope,
 ) : InternalDatabase {
     override val transactor: PsDatabase = PsDatabase(driver)
     override val queries: PowersyncQueries = transactor.powersyncQueries
+
+    // Could be scope.coroutineContext, but the default is GlobalScope, which seems like a bad idea. To discuss.
+    private val dbContext = Dispatchers.IO
```
kpgalligan, line 34: We should probably discuss this. Suspending calls that wind up calling the db will still block the thread when they try to access db calls. This would (in theory) be true even if SqlDelight's async mode was on. It never moved calls off of the calling thread, and the underlying sqlite calls aren't async. I picked `Dispatchers.IO` as it seems like an easy/obvious choice. `val scope: CoroutineScope` could be used, but I see that it's default is `GlobalScope`, which is something I am curious about. Why `GlobalScope`? In any case, I wouldn't launch local db calls with `GlobalScope`, so using `Dispatchers.IO`, but can discuss options.

DominicGBauer: It's unclear what the reason was for using GlobalScope. I agree that this is a better option.

DominicGBauer: @kpgalligan are there any potential downsides with this approach when thinking about the Swift SDK?

kpgalligan: Downsides with using `Dispatchers.IO` or `GlobalScope`?

Regardless of which, the changes I added move all real db calls to a different thread. Assuming Coroutines itself functions as expected, moving to a different thread shouldn't change anything with the public API. There's still the overall question of if sql-level errors should be catchable and, if so, how to encode them to be caught. SKIE suspend functions become throwable to Swift, but I believe that is only for `CancellationError` related to the async call. Other errors wouldn't be raised in a manner that is usable to a Swift dev. I haven't really focused on that, as except for the transaction block, the db calls haven't changed for what an external caller would expect, but it probably makes sense to formalize if sql/db should throw some kind of Swift error, and if so, have some definition of what that would look like.

But, internally moving to a different thread shouldn't impact the Swift side at all, except for the change in transaction block calls.

On `GlobalScope`, I wouldn't expect that to be useful to PowerSync. If PowerSync creates its own scope on startup, there really shouldn't be a reason to use `GlobalScope`. Now, whether the db calls should use `Dispatchers.IO` or something else, I don't have a hard opinion on that, but I do think the caller is relying on the PowerSync SDK to manage concurrency, and db calls on a background thread is generally recommended.

Of course, there is the critical init block when PowerSyncDatabaseImpl starts:
```
init {
        runBlocking {
            val sqliteVersion = internalDb.queries.sqliteVersion().executeAsOne()
            logger.d { "SQLiteVersion: $sqliteVersion" }
            checkVersion()
            logger.d { "PowerSyncVersion: ${getPowerSyncVersion()}" }
            internalDb.queries.powersyncInit()
            applySchema()
            updateHasSynced()
        }
    }
```

That must complete before anything else happens, which means it'll probably happen on the main thread for most apps. PowerSync isn't really unique here. It's been an app design issue forever. Avoiding that would be rather complex.

But I digress. Can hop on a call to chat about this.

DominicGBauer (approved):

PR #111, opened 2025-02-03
chore: update version to BETA 19

DominicGBauer:
## Description
Update to Beta 19 which will then include the SQL driver updates to allow to get a column by name using cursor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,22 @@
 
 ## 1.0.0-BETA19
 
-* Allow cursor to columns by name 
-* POTENTIAL BREAKING CHANGE: If you were using `SqlCursor` from SqlDelight previously for your own custom mapper then you must now change to SqlCursor exported by the PowerSyncSqlDelight module.
+* Allow cursor to columns by name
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.0.0-BETA19
+
+* Allow cursor to columns by name 
+* POTENTIAL BREAKING CHANGE: If you were using `SqlCursor` from SqlDelight previously for your own custom mapper then you must now change to SqlCursor exported by the PowerSyncSqlDelight module.
```
stevensJourney, line 6: Could you expand a bit on the `POTENTIAL` here?  It would be nice to know under which (some example) circumstances the changes are and are not breaking.

DominicGBauer: Added an example of the change that needs to be made

stevensJourney: Thanks for adding the example! Regarding the `POTENTIAL BREAKING CHANGE` phrasing—if this change requires users to update their code when upgrading, I think it should be labeled as a `BREAKING CHANGE` rather than softening it with "potential"

Since the impact is limited to users who were specifically using `SqlCursor` for custom mapping, the example in the description should be sufficient to clarify when action is needed.

This ensures it's clear that some users must take action while avoiding ambiguity.

DominicGBauer: Done

stevensJourney (approved):

PR #113, opened 2025-02-05
feat: add SqlCursor function improvements

DominicGBauer:
## Description

* Add additional SqlCursor functions `getStringOptional`, `getLongOptional`, `getDoubleOptional`, `getBooleanOptional` and `getBytesOptional` when using the column name as a parameter which return optional types. 
* `getString`, `getLong`, `getBytes`, `getDouble` and `getBoolean` used with the column name parameter will now throw an error for non-null values and expect a non-optional return type
* Throw errors for invalid columns on all cursor functions that use the column name parameter
* Update version to Beta20

Previously you would need to assert types and cast
```kotlin
return Todo(
  id: cursor.getString(name: "id")!,
  photoId: cursor.getString(name: "photo_id"),
  isComplete: cursor.getBoolean(name: "completed")! as! Bool,
)
```

now you can use this:

```kotlin
return Todo(
  id: cursor.getString(name: "id"),
  photoId: cursor.getStringOptional(name: "photo_id"),
  isComplete: cursor.getBoolean(name: "completed"),
)
```

## Testing
I've run all the demos and confirmed they work as they did previously.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 
+## 1.0.0-BETA20
+* Add cursor optional functions: `getStringOptional`, `getLongOptional`, `getDoubleOptional`, `getBooleanOptional` and `getBytesOptional` when using the column name which allow for optional return types
+* Throw errors for invalid column on all cursor functions
+* `getString`, `getLong`, `getBytes`, `getDouble` and `getBoolean` used with the column name will now throw an error for non-null values and expect a non optional return type
+
 ## 1.0.0-BETA19
 
-* Allow cursor to columns by name
+* Allow cursor to get values by column name e.g. `getStringOptional("id")`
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Changelog
 
+## 1.0.0-BETA20
+* Add cursor optional functions: `getStringOptional`, `getLongOptional`, `getDoubleOptional`, `getBooleanOptional` and `getBytesOptional` when using the column name which allow for optional return types
+* Throw errors for invalid column on all cursor functions
+* `getString`, `getLong`, `getBytes`, `getDouble` and `getBoolean` used with the column name will now throw an error for non-null values and expect a non optional return type
```
stevensJourney, line 6: Is this a breaking change from the previous `BETA19` release?

DominicGBauer: Yes. We haven't officially released BETA19 so I figured it did not require a mention.

stevensJourney (approved):

PR #124, opened 2025-02-18
Bucket priorities

simolus3:
This:

1. Refactors sync lines into sealed classes, making them easier to match in the streaming sync implementation.
2. Adds support for partial checkpoint complete messages, which we'll use to start a partial validation / checkpoint operation only involving some buckets.
3. Adds integratation tests that uses the complete client with only a mocked sync service. These assert that we're able to reat to partial checkpoint messages to reveal rows for partial sync operations.


core/src/commonMain/kotlin/com/powersync/db/SqlCursor.kt
```diff
@@ -19,6 +19,13 @@ public interface SqlCursor {
     public val columnCount: Int
 
     public val columnNames: Map<String, Int>
+
+    public fun asMap(): Map<String, Any?> =
```
stevensJourney, line 23: It looks like the function signature mentions this will return a `Map<String, Any?>` while the implementation seems like the map values will always be of type `String?`. Perhaps we should document the behaviour of this method?

simolus3: Sorry, I've used that for debugging but it shouldn't be part of this PR. I've removed the method now.

stevensJourney (approved):

PR #126, opened 2025-02-19
chore: use getBoolean in demo

DominicGBauer:
## Description
Use `getBoolean` in demo and remove unused import


stevensJourney (approved):

demos/supabase-todolist/shared/src/commonMain/kotlin/com/powersync/demos/powersync/Todo.kt
```diff
@@ -40,7 +40,7 @@ internal class Todo(
                 description = cursor.getString("description"),
                 createdBy = cursor.getStringOptional("created_by"),
                 completedBy = cursor.getStringOptional("completed_by"),
-                completed = cursor.getLongOptional("completed") == 1L,
+                completed = cursor.getBoolean( "completed"),
```
stevensJourney, line 43: This requires that the `completed` column should not be null (which can possible if we don't aren't strict in the demo). Do we currently always set the `completed column?

DominicGBauer: Yeah we set it to false initially

PR #127, opened 2025-02-20
Support Java 8

simolus3:
This:

1. Removes the use of Java toolchains for compiling code. These are ["rarely a good idea"](https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/) because newer JDKs are perfectly capable of emitting code for older JVMs (this is especially true for us since we only use Kotlin anyway).
2. Uses Java 8 as a target for the JVM-specific KMP targets. We're still targeting Java 17 for Android, like before.
3. Reformats code so that a clean `./gradlew build` works. I'll look into why formatting isn't applied on the CI, but that's less time-critical.
4. Upgrades stately to a version that supports Java 8.


DominicGBauer (dismissed):
Assuming you've tested this in the demo LGTM

DominicGBauer (approved):

PR #129, opened 2025-02-21
Simplify JVM build, start with integration tests

simolus3:
This:

1. Removes the native build for JVM targets. The build provided a compiled SQLite version and JNI bindings for update/commit/rollback hooks. But `org.xerial:sqlite-jdbc` (which we're using) also provides sqlite bindings, and also supports APIs for these hooks. So the native code is not actually required and can be omitted entirely.
2. Adds a new source set for integration tests. Integration tests will be able to actually use the database and sqlite core extensions. In this PR, I've set up integration tests for:
    - The JVM and JVM-8 targets: No special concerns here, since the JAR we build contains all the relevant native code. So we can combine unit and integration tests into a single target.
    - iOS simulator on armv8: We need to link our core extensions for this (we didn't do that previously, only our Swift SDK is combining the Kotlin + sqlite-core frameworks). I didn't add the framework as a library dependency (so we can keep the current Swift setup), but manually linking the framework into the Kotlin Native test executable is necessary. This allows combining unit and integration tests into a single target too.
    - Missing from this PR: Android will require a separate target for integration tests since our database factory requires access to the context.
3. Runs the linter and formatter in the CI.

As a follow-up to this, I think we can also remove our native code on Android with a refactoring (and set up integration tests for that too).


core/src/jvmMain/kotlin/com/powersync/PSJdbcSqliteDriver.kt
```diff
@@ -53,7 +53,7 @@ internal class PSJdbcSqliteDriver(
         listenersToNotify.forEach(Query.Listener::queryResultsChanged)
     }
 
-    private val connection: SQLiteConnection = DriverManager.getConnection(url, properties) as SQLiteConnection
+    val connection: SQLiteConnection = DriverManager.getConnection(url, properties) as SQLiteConnection
```
DominicGBauer, line 56: Maybe I missed where this is used outside of this but why does this not need to be private anymore?

DominicGBauer (commented):
LGTM just a minor question

core/src/jvmMain/kotlin/com/powersync/PSJdbcSqliteDriver.kt
simolus3: We're now accessing the connection in `DatabaseDriverFactory.jvm.kt` to listen for updates on the inner database connection (`PSJdbcSqliteDriver`) and forward it to the `PsSqlDriver`.

PR #130, opened 2025-02-24
Investigate failed gradle wrapper check

simolus3:
Trying to figure out why https://github.com/powersync-ja/powersync-kotlin/actions/runs/13493610812/job/37695985350 has failed.


simolus3:
And of course it works now... I'm not sure if that failure was a one-off or if there's an actual problem here then?

rkistner (approved):
Either way it's good to switch to the non-deprecated actions.

Issue #112, opened 2025-02-04
Crash on Android when using R8/Proguard


benoitletondor:
So on Android, release build are minified by R8/Proguard, it means that the code gets minified.

As a library, you can provide a proguard consumer rule file which describe which methods shouldn't be minified so that the full path will be present at runtime.

It looks like it's missing from Powersync and it leads to a crash:

```
 runtime.cc:708] Pending exception java.lang.NoSuchMethodError: no non-static method "Lcom/powersync/DatabaseDriverFactory;.onTransactionCommit(Z)V"
                                                                                                    runtime.cc:708]   at void io.requery.android.database.sqlite.SQLiteConnection.nativeExecute(long, long) (SourceFile:-2)
                                                                                                    runtime.cc:708]   at void io.requery.android.database.sqlite.SQLiteConnection.execute(java.lang.String, java.lang.Object[], D1.d) (SourceFile:31)
                                                                                                    runtime.cc:708]   at void io.requery.android.database.sqlite.SQLiteSession.endTransactionUnchecked(D1.d, boolean) (SourceFile:63)
                                                                                                    runtime.cc:708]   at void io.requery.android.database.sqlite.SQLiteSession.endTransaction(D1.d) (SourceFile:5)
                                                                                                    runtime.cc:708]   at void io.requery.android.database.sqlite.SQLiteDatabase.endTransaction() (SourceFile:9)
                                                                                                    runtime.cc:708]   at io.requery.android.database.sqlite.SQLiteDatabase io.requery.android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(boolean) (SourceFile:199)
                                                                                                    runtime.cc:708]   at io.requery.android.database.sqlite.SQLiteDatabase io.requery.android.database.sqlite.SQLiteOpenHelper.getWritableDatabase() (SourceFile:3)
                                                                                                    runtime.cc:708]   at z2.a io.requery.android.database.sqlite.SQLiteOpenHelper.getWritableDatabase() (SourceFile:1)
                                                                                                    runtime.cc:708]   at java.lang.Object G.i.invoke() (SourceFile:201)
                                                                                                    runtime.cc:708]   at java.lang.Object B9.r.getValue() (SourceFile:21)
                                                                                                    runtime.cc:708]   at z2.a M4.l.c() (SourceFile:5)
                                                                                                    runtime.cc:708]   at java.lang.Object Z2.f.invoke() (SourceFile:5)
[...]

PR #132, opened 2025-02-26
Android: Support minification with ProGuard and R8

simolus3:
On Android, a build step crawls the call graph to minify method names and remove unused methods. This step uses tools like R8 (or ProGuard on much older SDKs) that only look at Java bytecode to determine which methods are called. In our SDK, there are two restrictions that these tools miss:

1. `com.powersync.DatabaseDriverFactory.setupSqliteBinding` is a native method implemented with JNI. Its name must not be changed for the linking to work.
2. In the same class, `onTableUpdate` and `onTransactionCommit` are called from C code with JNI. If the methods are renamed or removed, we'd crash.

For these reasons, we need to add ProGuard files marking these methods as used and unsuitable for minifying.

To test this, I've copied some of the simpler integration tests on the `core` project into a `core-tests-android` where they can run on an emulator / real device (due to limitations in the Kotlin multiplatform and Android Gradle platforms, it doesn't appear to be possibl possible to share this code across platforms). The tests in that project started passing after adding the ProGuard rules.

Closes https://github.com/powersync-ja/powersync-kotlin/issues/112.


benoitletondor:
This doesn't close #112 , check https://github.com/powersync-ja/powersync-kotlin/issues/112#issuecomment-2687511047 for details

Issue #135, opened 2025-03-01
Race condition in `PsSqlDriver`?


ecstatic-morse:
I'm seeing intermittent crashes inside [``PsSqliDriver.fireTableUpdates`](https://github.com/powersync-ja/powersync-kotlin/blob/738fd0a041c43e1efc0374fbf6357ddc504c71ed/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt#L36), e.g.

<details>
<summary>Stack trace</summary>
```
     Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=7; index=7
       at java.util.LinkedHashMap.keysToArray(LinkedHashMap.java:703)
       at java.util.LinkedHashMap.keysToArray(LinkedHashMap.java:691)
       at java.util.HashSet.toArray(HashSet.java:376)
       at java.util.ArrayList.<init>(ArrayList.java:188)
       at kotlin.collections.CollectionsKt___CollectionsKt.toMutableList(_Collections.kt:1343)
       at kotlin.collections.CollectionsKt___CollectionsKt.toList(_Collections.kt:1324)
       at com.powersync.PsSqlDriver.fireTableUpdates(PsSqlDriver.kt:37)
       at com.powersync.DatabaseDriverFactory.onTransactionCommit(DatabaseDriverFactory.android.kt:29)
       at io.requery.android.database.sqlite.SQLiteConnection.nativeExecuteForChangedRowCount(SQLiteConnection.java)
       at io.requery.android.database.sqlite.SQLiteConnection.executeForChangedRowCount(SQLiteConnection.java:778)
       at io.requery.android.database.sqlite.SQLiteSession.executeForChangedRowCount(SQLiteSession.java:764)
       at io.requery.android.database.sqlite.SQLiteStatement.executeUpdateDelete(SQLiteStatement.java:71)
       at app.cash.sqldelight.driver.android.AndroidPreparedStatement.execute(AndroidSqliteDriver.kt:261)
       at app.cash.sqldelight.driver.android.AndroidSqliteDriver$execute$2.invoke(AndroidSqliteDriver.kt:184)
       at app.cash.sqldelight.driver.android.AndroidSqliteDriver$execute$2.invoke(AndroidSqliteDriver.kt:184)
       at app.cash.sqldelight.driver.android.AndroidSqliteDriver.execute-zeHU3Mk(AndroidSqliteDriver.kt:169)
[...]

PR #136, opened 2025-03-03
Improve Watch Queries

stevensJourney:
# Overview

This PR improves the watched query logic by optimizing table update handling, enhancing thread safety, and simplifying the implementation.

## Table Update Optimization

- Table changes in a transaction are now tracked using a `Set`, preventing redundant entries. Since SQLite emits an update per row, the previous implementation could accumulate large arrays of duplicate table changes.

## Thread Safety for Table Updates

- Resolves [issue #135](https://github.com/powersync-ja/powersync-kotlin/issues/135), where the current driver’s pending update storage was not thread-safe.
- The pending table changes set is now protected with Mutex locks.

## Implementation Cleanup

- Replaces the existing SQLDelight watched query interface with a simpler, Flow-based solution.
- The new approach leverages the internal database’s `getAll` method instead of the driver’s synchronous methods.
- Table change flow logic has been refactored for better debouncing, allowing custom `throttleMs` values per query.

## Removal of Blocking Calls

- Previously, the watch query API synchronously determined the tables to watch, which impacted performance (e.g., in the Supabase Todo List app). This logic now uses the asynchronous getAll method, improving efficiency.

## Testing

This was tested locally using both the Kotlin and Swift SDKs.

https://github.com/user-attachments/assets/17350fdb-367b-4987-916c-e3eb540dc911

## Swift

Related Swift PR https://github.com/powersync-ja/powersync-swift/pull/26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.0-BETA27
+
+* Improved watch query internals. Added the ability to throttle watched queries.
+
 ## 1.0.0-BETA26
 
 * Support bucket priorities and partial syncs.
--- a/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
@@ -18,7 +18,7 @@ internal class PsSqlDriver(
     private val tableUpdatesFlow = MutableSharedFlow<Set<String>>(replay = 0)
 
     // In-memory buffer to store table names before flushing
-    private val pendingUpdates =  AtomicMutableSet<String>()
+    private val pendingUpdates = AtomicMutableSet<String>()
 
     fun updateTable(tableName: String) {
         scope.launch {
@@ -33,16 +33,13 @@ internal class PsSqlDriver(
     }
 
     // Flows on table updates containing tables
-    fun updatesOnTables(tableNames: Set<String>): Flow<Unit> = tableUpdatesFlow.asSharedFlow().filter { it.intersect(tableNames).isNotEmpty() }.map { }
+    // The table names here are raw SQLite tables
+    fun updatesOnTables(tableNames: Set<String>): Flow<Unit> =
+        tableUpdatesFlow.asSharedFlow().filter { it.intersect(tableNames).isNotEmpty() }.map { }
 
-    suspend fun fireTableUpdates() {
-        val updates = pendingUpdates.toSet()
-        pendingUpdates.clear()
-
-        if (updates.isEmpty()) {
-            return
+    fun fireTableUpdates() {
+        scope.launch {
+            tableUpdatesFlow.emit(pendingUpdates.toSet(true))
         }
-
-        tableUpdatesFlow.emit(updates)
     }
 }
--- a/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
@@ -33,9 +33,21 @@ internal class PsSqlDriver(
     }
 
     // Flows on table updates containing tables
-    // The table names here are raw SQLite tables
-    fun updatesOnTables(tableNames: Set<String>): Flow<Unit> =
-        tableUpdatesFlow.asSharedFlow().filter { it.intersect(tableNames).isNotEmpty() }.map { }
+    fun updatesOnTables(tableNames: Set<String>): Flow<Unit> {
+        // Spread the input table names in order to account for internal views
+        val resolvedTableNames =
+            tableNames
+                .flatMap { t -> setOf("ps_data__$t", "ps_data_local__$t", t) }
+                .toSet()
+        return tableUpdatesFlow
+            .asSharedFlow()
+            .filter {
+                it
+                    .intersect(
+                        resolvedTableNames,
+                    ).isNotEmpty()
+            }.map { }
+    }
 
     fun fireTableUpdates() {
         scope.launch {
--- a/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
@@ -2,6 +2,7 @@ package com.powersync
 
 import app.cash.sqldelight.db.SqlDriver
 import com.powersync.utils.AtomicMutableSet
+import com.powersync.utils.throttle
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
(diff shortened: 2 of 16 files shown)


core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
```diff
@@ -1,46 +1,49 @@
...
     private val driver: SqlDriver,
     private val scope: CoroutineScope,
 ) : SqlDriver by driver {
     // MutableSharedFlow to emit batched table updates
-    private val tableUpdatesFlow = MutableSharedFlow<List<String>>(replay = 0)
+    private val tableUpdatesFlow = MutableSharedFlow<Set<String>>(replay = 0)
 
     // In-memory buffer to store table names before flushing
-    private val pendingUpdates = mutableSetOf<String>()
+    private val pendingUpdates = AtomicMutableSet<String>()
```
simolus3, line 20: I generally subscribe to the idea that the use synchronized collections is an anti-pattern because it's still possible to have concurrency issues with then when not used properly. That's not the case here, but I still think we should remove methods from `AtomicMutableSet` that we're not using.

So perhaps we can remove the unused `remove`, `clear` and `contains` methods? And IMO the `clear` parameter on `toSet` can be removed too, maybe just call the method `toSetAndClear`. And then we could rename the class to e.g. `BatchBuilder` or something, but that's a nit and might be going a bit far.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -233,8 +238,9 @@ internal class PowerSyncDatabaseImpl(
     override fun <RowType : Any> watch(
         sql: String,
         parameters: List<Any?>?,
+        throttleMs: Long?,
```
simolus3, line 241: FWIW I prefer using a `Duration` type on languages that have them in their stdlib, but it looks like we're already using milliseconds as `Long` elsewhere and it's probably better to be consistent.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
```diff
@@ -189,21 +160,31 @@ internal class InternalDatabaseImpl(
...
+
+            updatesOnTables()
+                // onSubscription here is very important.
+                // This ensures that the initial result and all updates are emitted.
+                .onSubscription {
+                    send(getAll(sql, parameters = parameters, mapper = mapper))
+                }.filter {
+                    // Only trigger updates on relevant tables
+                    it.intersect(tables).isNotEmpty()
+                }.throttle(throttleMs ?: DEFAULT_WATCH_THROTTLE_MS)
```
simolus3, line 183: Maybe add a comment here why `throttle` is fine despite possibly dropping events when `getAll` is too slow (in my understanding, it's because we only need to ensure that an update after `getAll` started will cause another emission later, and that's guaranteed because we're keeping the latest event).

core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
```diff
@@ -1,46 +1,49 @@
...
+
+    // Flows on any table change
+    // This specifically returns a SharedFlow for downstream timing considerations
+    fun updatesOnTables(): SharedFlow<Set<String>> =
+        tableUpdatesFlow
+            .asSharedFlow()
+
+    suspend fun fireTableUpdates() {
+        // Use the same scope as the async table updates, this should help with queuing
+        val job =
```
simolus3, line 42: What's the purpose of wrapping this in a job that's immediately awaited?

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
```diff
@@ -189,21 +160,31 @@ internal class InternalDatabaseImpl(
...
-        ).asFlow().mapToList(scope.coroutineContext)
-    }
+    ): Flow<List<RowType>> =
+        // Use a channel flow here since we throttle (buffer used under the hood)
+        // This causes some emissions to be from different scopes.
+        channelFlow {
+            // Fetch the tables asynchronously with getAll
+            val tables =
+                getSourceTables(sql, parameters)
+                    .filter { it.isNotBlank() }
```
simolus3, line 172: Do we still need this given that the `ps_data__` stripping is gone?

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
```diff
@@ -189,21 +160,31 @@ internal class InternalDatabaseImpl(
...
+                    .toSet()
+
+            updatesOnTables()
+                // onSubscription here is very important.
+                // This ensures that the initial result and all updates are emitted.
```
simolus3, line 177: Is this any different to just moving the `send` out like this? The `channelFlow` is cold too, so it would run for each subscriber.

```Kotlin
send(getAll(sql, parameters = parameters, mapper = mapper))
updatesOnTables()
  .filter { ... }
  .throttle()
  .collect { sendAll(...) }
```

I think the important bit is that we need to schedule another emission if the initial query and an update on a watched table is happening concurrently.
But `onSubscription` doesn't appear to do that, looking at its implementation it runs the callback first and then subsribes to the remaining flow.

simolus3 (commented):
Thanks! I'm happy with the refactoring, and most of these comments are nits.

I have one concern about a possible race condition in the `watch` implementation, e.g. what if:

1. Thread `A` subscribes to a `watch()` flow and starts running the initial query in `onSubscribe`.
2. Before the query completes, thread B writes to a table affected by that watch (which should be possible with WAL).
3. The initial `send` completes with the old data.
4. Only after `onSubscribe` completes are we listening to table changes, so we don't get notified about the concurrent write and don't fetch again, like we should.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
simolus3: `onSubscribe` is actually fine as discussed offline.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
stevensJourney: I'm not entirely sure why this method returns potentially blank table names, but I think it still makes sense to filter blank strings from this list.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
stevensJourney: I agree, this is for consistency.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
stevensJourney: Good point. I added a comment.

core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
stevensJourney: Discussed offline. Updating the mutable set to use the `atomicfu` `SynchronizedObject` is a better approach and removes the need for this.

core/src/commonMain/kotlin/com/powersync/PsSqlDriver.kt
stevensJourney: Good suggestion! I've made some of these changes when updating to a synchronised object implementation.

core/src/commonMain/kotlin/com/powersync/db/internal/InternalDatabaseImpl.kt
simolus3: Yeah I don't think `sqlite_master.name` can be empty, but it also doesn't hurt so we can leave it is.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.0-BETA27
```
simolus3, line 3: If we want to release this after merging (without another PR to bump the version), we should also update `LIBRARY_VERSION` in `gradle.properties`. `docs/Release.md` describes the steps necessary for a release afterwards.

simolus3 (approved):
LGTM!

CHANGELOG.md
stevensJourney: 👍  There is another open issue for SyncStatus updates not reflecting the downloading and uploading updates. I'd like to batch those fixes in a single release.

PR #143, opened 2025-03-12
Demo: Add example using a foreground service for synchronization 

simolus3:
In the `supabase-todolist-demo` project, this adds a new Android app that will manage the PowerSync connection task in a service. I've put this into a new app project to avoid too many changes to the existing project (it complicates the sync setup a bit) while still being able to reuse UI code with the existing demo (so we can keep maintaining this example easily).

In the shared app, I've introduced [Koin](https://insert-koin.io/) for dependency injection. I think it's a good fit for that demo because it suppports multiplatform (with very good support for Android too). We have to stop creating the `PowerSyncDatabase` in a `remember` block of the compose component because we want to use the same instance outside of it as well. So now we use Koin and inject the database into the UI component (and into the background sync service in the new Android app).

The only other change to the shared app is that we don't unconditionally call `connect()` and `disconnect()` in the auth view model anymore. The view model will still do that by default, but this behavior is disabled by the background sync app.

Apart from that, the only change is starting a foreground service once the user is logged in. This keeps the app process (and thus, the connector + the streaming sync implementation) alive, allowing us to keep the synchronization active while the main app is closed.
It looks like there's almost nothing we could do in the SDK to make this easier. Users create a singleton of their database with whatever DI framework they're using, and then they can connect in a background service without much additional setup.

### Demo video

https://github.com/user-attachments/assets/ac031555-c1bc-463a-ab80-456f1160e310

After closing the app and removing it from the recently used apps (destroying the activity), the sync client keeps running. I can change a row in supabase while the app is closed, turn on airplane mode and open the app - the changes were already synced!

I've also tested this on an Android 9 emulator.


demos/supabase-todolist/README.md
```diff
@@ -58,3 +58,11 @@ SUPABASE_ANON_KEY=foo
...
 
 Choose a run configuration for the Android or iOS target in Android Studio and run it.
+
+For Android, this demo contains two Android apps:
+
+- [`androidApp/`](androidApp/): This is a regular compose UI app using PowerSync.
+- [`androidBackgroundSync/`](androidBackgroundSync/): This example differs from the regular app in
+  that it uses a foreground service managing the synchronization process. The service is started
+  in the main activity and keeps running even after the app is closed.
+  For more notes on background sync, see [this document](docs/BackgroundSync.md).
```
stevensJourney, line 68: Could we add a link to the background sync docs document here?

stevensJourney: Sorry I missed the link is already there

stevensJourney (approved):
Happy with the changes from my side.

PR #145, opened 2025-03-17
Concurrent Connections

stevensJourney:
# Overview

This PR introduces minor improvements to the SQLite driver functionality.

## Unified Concurrency
The PowerSync client now supports readLock and writeLock APIs, aligning with the functionality in other SDKs.

Previously, concurrency was implemented differently across the various platform SQLite drivers. The Android and iOS drivers had their own concurrency models. The iOS driver opened a single read and write connection, while the Android driver’s read connections were dependent on available system memory ("The maximum number of connections used for parallel queries depends on the device’s memory and possibly other factors.").

The JVM driver only created a single connection, preventing concurrent reads.

The current driver interface required substantial work on the driver implementation to route queries to the appropriate internal connection. Implementing concurrency for the JVM driver would have been complex. Instead, we unified the concurrency model across all platforms by modifying the implementation to use multiple SQLite drivers.

Now, a single write connection driver is used for write operations, while a pool of read connection drivers is maintained.

### Android
The current Android SQLite driver uses [Requery](https://github.com/requery/requery), which does not support SQLite table update hooks in its API. Although PRs for adding SQLite hooks were introduced recently, they have not been released yet.

We currently use custom native C interops to access SQLite hooks via C APIs. However, these interops are limited and only function properly with a single SQLite connection/driver.

Implementing update hooks for multiple connections in native C interops is complex. As a result, this PR switches the Android driver to use the [Xerial JDBC SQLite](https://github.com/xerial/sqlite-jdbc) library (which is also used in the JVM build). This library supports SQLite hooks as part of its API.

Although the Xerial library requires shipping JNI binaries with our SDK, the Gradle build scripts have been updated to include the necessary JNI libraries during the build process.

The JDBC driver code is now shared between the Android and JVM targets, simplifying the implementation.

Removing the Requery dependency also streamlines installation, as users no longer need to add the Jitpack Maven repository to their build scripts.

TODOs:

- [x]  Update documentation and demos to reflect the removal of the Requery dependency.

### iOS
The iOS driver currently uses [SQLiteR](https://github.com/touchlab/SQLiter), which interfaces closely with the SQLite C APIs. However, SQLiteR does not expose SQLite hooks in its API, requiring us to use a native C interop for hook functionality.

The current hook implementation only works with the last driver created by the DatabaseDriverFactory. This implementation has been updated to support multiple connections.

Previously, the iOS driver loaded the PowerSync SQLite core extension statically by calling sqlite3_auto_extension through a native C interop. This PR now dynamically loads the extension and throws exceptions if the PowerSync core framework cannot be located.

Full-text search is now enabled by default for iOS.

## Native Interops
Much of the native interop code has been removed, as it is no longer necessary.

## Database Options
The Android and JVM drivers now support opening a database using a dbDirectory parameter. This feature is not supported on iOS.

## SQLDelight Queries

We currently use SQLDelight's dialect for internal PowerSync queries. These operations use the driver directly. Which easily can cause synchronous code execution to block the main thread (some examples of this have been removed). Using the driver directly also has some slight chance of interfering with connection locks. 

The use of SQLDelight dialect internally has mostly been removed and replaced with raw SQLite queries.


core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
```diff
@@ -0,0 +1,152 @@
...
+    @Suppress("UNCHECKED_CAST")
+    suspend fun <R> withConnection(action: suspend (connection: TransactorDriver) -> R): R {
+        if (closed.value) {
+            throw PowerSyncException(
+                message = "Cannot process connection pool request",
+                cause = Exception("Pool is closed"),
+            )
+        }
+
+        val wrappedDeferred = DeferredAction(CompletableDeferred<Any?>(), action)
```
simolus3, line 45: I think this might have the downside that `action` might be executed in a different coroutine scope, breaking cancellations? I'm not sure about this though.

core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
```diff
@@ -0,0 +1,152 @@
...
+) {
+    var busy: Boolean = false
+}
+
+internal data class DeferredAction<R>(
+    val deferred: CompletableDeferred<R>,
+    val action: suspend (connection: TransactorDriver) -> R,
+)
+
+internal class ConnectionPool(
```
simolus3, line 24: I think a _really_ cool building block for this functionality are [channels](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/). Channels are fair and provide both fan-in and fan-out, which can be used to express this safely. A setup could be that:

1. We use a single rendezvous channel for `ConnectionPool`.
2. We spawn `size` jobs in `scope`. Every job will create a driver and have a `while (true)` loop that posts its connection and a `CompletableDeferred` to the channel.
3. When we want to use a connection, we `receive()` from the channel to obtain the connection and the deferred to complete when we're done with it.
4. To close the pool, we `cancel()` the channel. This will interrupt pending waiters on `receive()` and cause the sending jobs to get shut down too. So we just need to join them and can be sure that all resources are cleaned up.

I didn't test this, but I imagine something like this could work. This should have the same semantics as the current implementation while delegating most of the internal synchronization logic to channels:

```Kotlin
internal class ConnectionPool(
    factory: () -> PsSqlDriver,
    size: Int = 5,
) {
    private val available = Channel<Pair<TransactorDriver, CompletableDeferred<Unit>>>()
    private val connections: List<Job> = buildList {
        repeat(size) {
            scope.launch {
                val driver = TransactorDriver(factory())
                try {
                    while (true) {
                        val done = CompletableDeferred<Unit>()
                        try {
                            available.send(driver to done)
                        } catch (_: ClosedSendChannelException) {
                            break // Pool closed
                        }

                        done.await()
                    }
                } finally {
                    driver.driver.close()
                }
            }
        }
    }

    suspend fun <R> withConnection(action: suspend (connection: TransactorDriver) -> R): R {
        val (connection, done) = try {
            available.receive()
        } catch (e: PoolClosedException) {
            throw PowerSyncException(
                message = "Cannot process connection pool request",
                cause = e,
            )
        }

        try {
            return action(connection)
        } finally {
            done.complete(Unit)
        }
    }

    suspend fun close() {
        available.cancel(PoolClosedException)
        connections.joinAll()
    }
}

internal object PoolClosedException : CancellationException("Pool is closed")
```

core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
stevensJourney: Yeah, cancellations would have been an issue. I've updated the pooling implementation to the recommended channel approach which should not have this issue.

build.gradle.kts
```diff
@@ -27,9 +27,6 @@ allprojects {
         maven("https://cache-redirector.jetbrains.com/intellij-dependencies")
         // Repo for the backported Android IntelliJ Plugin by Jetbrains used in Ultimate
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies/")
-        maven("https://jitpack.io") {
-            content { includeGroup("com.github.requery") }
-        }
```
simolus3, line 32: Are we suggesting somewhere on the docs that users do this as well? I couldn't find anything from a quick search, but if we are we should update that too.

core/README.md
```diff
@@ -1,19 +1,31 @@
...
+- `jvmMain` - JVM specific code which includes an implementation of `DatabaseDriverFactory`.
+- `iosMain` - iOS specific code, which includes am implementation of `DatabaseDriverFactory` class
+  that creates an instance of `app.cash.sqldelight.driver.native.NativeSqliteDriver` and also sets
+  up native SQLite bindings for iOS.
 
 ## Note on SQLDelight
 
-The PowerSync core module, internally makes use of [SQLDelight](https://cashapp.github.io/sqldelight) for it database API and typesafe database query generation.
+The PowerSync core module, internally makes use
+of [SQLDelight](https://cashapp.github.io/sqldelight) for it database API and typesafe database
```
simolus3, line 27: While we're at it, it looks like the current URL is
```suggestion
of [SQLDelight](https://sqldelight.github.io/sqldelight/latest/) for it database API and typesafe database
```

core/build.gradle.kts
```diff
@@ -214,6 +219,9 @@ kotlin {
         val commonJava by creating {
             kotlin.srcDir("commonJava")
```
simolus3, line 220: Looks like I might have been the one smuggling that in, but I think this `srcDir` is not necessary.

core/build.gradle.kts
```diff
@@ -296,22 +303,20 @@ android {
...
         }
     }
+    ndkVersion = "27.1.12297006"
+}
 
-    externalNativeBuild {
-        cmake {
-            path = project.file("src/androidMain/cpp/CMakeLists.txt")
+androidComponents.onVariants { variant ->
+    if (variant.buildType == "debug" || variant.buildType == "release") {
```
simolus3, line 317: Are there other variants where this shouldn't apply?

core/proguard-rules.pro
```diff
@@ -9,3 +9,5 @@
     private void onTableUpdate(java.lang.String);
     private void onTransactionCommit(boolean);
 }
+-keep class org.sqlite.** { *; }
+-dontwarn java.sql.JDBCType
```
simolus3, line 13: I wonder if we can be more specific here, but doing this appears to be the common approach for `org.sqlite`. If you feel like it, you could try only keeping the classes [mentioned here](https://github.com/xerial/sqlite-jdbc/blob/bd0e99541a70a31da41f7d7c4153e7ac62840cad/src/main/java/org/sqlite/core/NativeDB.c#L443-L498). But if it's too much effort or breaks the tests, I'd say don't worry.

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
```diff
@@ -97,7 +213,7 @@ class DatabaseTest {
                 try {
                     database.writeTransaction {
                         it.execute("DELETE FROM users;")
-                        it.execute("syntax error, revert please")
+                        it.execute("syntax error, revert please (this is intentional from the unit test)")
```
simolus3, line 216: This change sounds like it took some time to figure out where that error came from at some point. Sorry for the original text 🙈

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
```diff
@@ -114,6 +230,71 @@ class DatabaseTest {
...
+            // Close the database. This should close the read pool
+            // The pool should wait for jobs to complete before closing
+            val closeJob =
+                async {
+                    database.close()
+                }
+
+            // Wait a little for testing
+            // Spawns in a different context for the delay to actually take affect
+            async { withContext(Dispatchers.Default) { delay(500) } }.await()
```
simolus3, line 261: Not a huge fan of this, but I can't see a much better way either. We could add this to `ConnectionPool`:

```Kotlin
    @OptIn(DelicateCoroutinesApi::class)
    internal val closeRequested: Boolean get() = available.isClosedForReceive
```

But when we'd still have to expose the pool on `InternalDatabaseImpl` and the database impl on `PowerSyncDatabaseImpl` as `internal` fields to access them in this test (to poll on `closeRequested` becoming `true`).

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
```diff
@@ -114,6 +230,71 @@ class DatabaseTest {
...
+
+            // The database should not close yet
+            assertEquals(actual = database.closed, expected = false)
```
simolus3, line 264: Can we assert that, in this state, calling `database.readLock` again throws?

core/src/commonJava/kotlin/com/powersync/db/JdbcSqliteDriver.kt
```diff
@@ -118,34 +93,49 @@ internal class PSJdbcSqliteDriver(
...
     }
 
-    internal fun enableWriteAheadLogging() {
-        val executed = connection.prepareStatement("PRAGMA journal_mode=WAL;").execute()
-        check(executed) { "journal_mode=WAL failed" }
+    private inner class Transaction(
+        override val enclosingTransaction: Transaction?,
+    ) : Transacter.Transaction() {
+        init {
+            connection.prepareStatement("BEGIN TRANSACTION").use(PreparedStatement::execute)
```
simolus3, line 114: This doesn't support nested transactions. I think we should either:

1. Keep some kind of depth counter on this `Transaction` class to emit `SAVEPOINT` / `RELEASE` / `ROLLBACK TO` statements on inner transactions.
2. Assert that `enclosingTransaction` is null, because ending this inner transaction would otherwise and the enclosing one?

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -89,17 +97,22 @@ internal class PowerSyncDatabaseImpl(
     private var syncJob: Job? = null
     private var uploadJob: Job? = null
 
-    init {
-        val res = ActiveDatabaseGroup.referenceDatabase(logger, identifier)
-        resource = res.first
-        clearResourceWhenDisposed = res.second
+    // This is set in the init
+    private var powerSyncVersion: String? = null
```
simolus3, line 101: I suppose the typical pattern for this in Kotlin would be to

```suggestion
    private lateinit var powerSyncVersion: String
```

And then remove the null-checks when accessing it.

core/src/commonMain/kotlin/com/powersync/db/internal/PowerSyncTransaction.kt
```diff
@@ -1,34 +1,8 @@
...
-    ): List<RowType>
-
-    @Throws(PowerSyncException::class, CancellationException::class)
-    public fun <RowType : Any> get(
-        sql: String,
-        parameters: List<Any?>? = listOf(),
-        mapper: (SqlCursor) -> RowType,
-    ): RowType
-}
+public class PowerSyncTransactionImpl(
```
simolus3, line 5:
```suggestion
internal class PowerSyncTransactionImpl(
```

core/src/iosSimulatorArm64Test/kotlin/com/powersync/testutils/TestUtils.iosSimulatorArm64.kt
```diff
@@ -13,3 +13,5 @@ actual fun cleanup(path: String) {
         SystemFileSystem.delete(resolved)
     }
 }
+
+actual fun getTempDir(): String? = null
```
simolus3, line 17: Maybe add a comment explaining that while we could use `SystemTemporaryDirectory` here, we're deliberately not doing that because SQLiter doesn't support setting a directory and that tests would attempt to do so if `getTempDir()` were to return something.

static-sqlite-driver/build.gradle.kts
```diff
@@ -0,0 +1,189 @@
...
+import org.jetbrains.kotlin.konan.target.PlatformManager
+
+plugins {
+    alias(libs.plugins.kotlinMultiplatform)
+    alias(libs.plugins.downloadPlugin)
+    id("com.powersync.plugins.sonatype")
+}
+
+val sqliteVersion = "3450200"
+val sqliteReleaseYear = "2024"
```
simolus3, line 19: When adding this originally, I've copied the version from the previous `:core` build definition.

Any reason for not updating `sqlite-jdbc` and this to `3.49.1`?

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
simolus3: Come to think of it, it's weird that `database.closed` is false but we can't use the database because it's closing. I'm not sure if there should be a `isClosing` getter in addition to `closed` or if it doesn't matter.

simolus3 (commented):
This generally looks really nice, it's awesome to see how much custom native code we can throw out with this.

I have a bunch of comments, but they're all pretty minor.

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
stevensJourney: 😄 I just added this for some clarity, it didn't take too long to see the log was coming from this test.

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
stevensJourney: Discussed offline. This delay here is to ensure that the database did not close immediately after the `close` method was called since there is an active readLock open. The close method should wait for the lock to complete.  
We currently only mark the database as `closed=true` after the database is actually closed. During this state some operations might succeed while others will fail due to the underlaying database being closed.

core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
stevensJourney: I've added a check to assert that new calls to readLock fail while the database is closing. 
The concept of an isClosing state is interesting. I think we can add that if there are any issues in practice. For now some read lock requests might acquire a lock during the initial phase of closing. In which case the close operation will wait for that lock operation to complete.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
stevensJourney: That is nicer! I've implemented this.

static-sqlite-driver/build.gradle.kts
stevensJourney: I could not find any reason not to update this. I originally used the same JDBC version as the core. I've updated all the versions now to `3.49.1`. Tests pass on all the platforms including the Swift SDK.

core/src/iosSimulatorArm64Test/kotlin/com/powersync/testutils/TestUtils.iosSimulatorArm64.kt
stevensJourney: Going point! I've added a comment to explain this.

core/build.gradle.kts
stevensJourney: No problem. I've removed this.

core/src/commonJava/kotlin/com/powersync/db/JdbcSqliteDriver.kt
stevensJourney: We currently set `noEnclosing = true` whenever we start a transaction, but I have also now added an assertion in the driver.

core/proguard-rules.pro
simolus3: Alright, then let's keep it as is. Probably better anyway because we'd have to maintain these definitions for possible future updates to that package.

simolus3 (dismissed):
LGTM 🚀

simolus3 (approved):

PR #147, opened 2025-03-19
Link SQLite statically on native platforms

simolus3:
All our client SDKs need the PowerSync core extension to be loaded to work. On Apple platforms however, SQLite extension loading is disabled. This means that our SDK must not link the SQLite library from the platform, and instead link a custom version compiled with extension support.

This:

1. Removes the last remnants of native builds in the `:core` project. We don't need it for Android since we're now using sqlite-jdbc. We also don't need it for iOS since we can call the cinterop methods exposed from the SQLiter dependency.
2. Disables linking SQLite with the option provided by SQLDelight (since that one uses the one from the system).
3. Adds a new project whose only purpose is to compile `sqlite3` into a static archive. This archive is then included on a cinterop definition. This makes the Kotlin compiler bundle the archive with the klib, dependents will automatically link it. 
4. Depend on the new project from `:persistence`. This means that we now link SQLite statically.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -56,27 +56,17 @@ val unzipSQLiteSources by tasks.registering(Copy::class) {
     into(sqliteSrcFolder)
 }
 
-val buildCInteropDef by tasks.registering {
+val compileSqliteForTesting by tasks.registering(Exec::class) {
     dependsOn(unzipSQLiteSources)
 
-    val interopFolder =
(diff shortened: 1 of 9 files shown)


stevensJourney (approved):
Nice work!!!! i tested this for both debug and release builds of the Supabase Compose demo. I confirmed a debug build of the Swift demo also works.

PR #148, opened 2025-03-21
Refactor mutexes and duplicate instance detection logic

simolus3:
This simplifies the recently-added logic to warn for duplicate databases. The actual functionality is unchanged, but:

1. The detection for duplicate databases and the mutex allocation to guard against two active syncs was unified into a single class (`ActiveDatabaseGroup`). 
2. On `PowerSyncDatabase`, also guard `disconnect()` with a mutex. Since connecting, disconnecting and closing the database are all operations that can interfere with each other, I've used a single mutex for all of them.
3. On native platforms, use a cleaner to automatically remove databases from the duplicate instance count when they're no longer referenced without `close()` being called. This can avoid a small memory leak. We should also investigate doing that on Java platforms, but convenient APIs are only available from Java 9+. So maybe something for another time...


stevensJourney (approved):
Happy with the changes!

Issue #150, opened 2025-03-23
Question: Is it possible to have a permanent local-only table mixed with synced tables?


luanrodrigsr:
Hello!

I tried to implement a local-only table to insert data that will stay  locally saved forever (until the user decides to clean up the database). But i'm seeing that all changes inserted are being enqueued to be send to my backend.

Database schema:
```Kotlin
Table.localOnly(
	name = "produtos",
	columns = listOf(
		Column.text(Keys.DESCRICAO),
                Column.text(Keys.CODIGO),
                Column.text(Keys.PRECO),
                Column.text(Keys.UPDATED_AT),
                Column.text(Keys.UPDATED_BY_ID),
                Column.text(Keys.TENANT_ID),
                Column.integer(Keys.ARCHIVED),
	),
)
```

Logcat:
```
2025-03-23 00:21:20.347 32627-610   System.out              co.sprinte.gst                       I  message response: Request failed... input must be a 24 character hex string, 12 byte Uint8Array, or an integer
2025-03-23 00:21:20.348 32627-610   Database                co.sprinte.gst                       I  [CrudTransaction::complete] Completing transaction with checkpoint null
2025-03-23 00:21:20.387 32627-610   System.out              co.sprinte.gst                       I  DatabaseRequestData(table=produtos, data={codigo=7500435115827, descricao=Sabao Liq Ariel Conc, preco=28,90, id=7500435115827})
2025-03-23 00:21:20.579 32627-609   System.out              co.sprinte.gst                       I  message response: Request failed... input must be a 24 character hex string, 12 byte Uint8Array, or an integer
2025-03-23 00:21:20.579 32627-609   Database                co.sprinte.gst                       I  [CrudTransaction::complete] Completing transaction with checkpoint null
2025-03-23 00:21:20.628 32627-618   System.out              co.sprinte.gst                       I  DatabaseRequestData(table=produtos, data={codigo=7500435116305, descricao=Condicionador Panten, preco=29,90, id=7500435116305})
2025-03-23 00:21:20.849 32627-609   System.out              co.sprinte.gst                       I  message response: Request failed... input must be a 24 character hex string, 12 byte Uint8Array, or an integer
2025-03-23 00:21:20.850 32627-609   Database                co.sprinte.gst                       I  [CrudTransaction::complete] Completing transaction with checkpoint null
[...]

PR #154, opened 2025-03-27
Update Schema

stevensJourney:
# Overview

https://github.com/powersync-ja/powersync-kotlin/issues/150 recently raised an issue with local only tables. 

This was caused by incorrect serialisation of the provided `Schema`. Specifically, the JSON payload for the schema supplied camel case field names for `localOnly`, `insertOnly` and `viewName` instead of the required `local_only`, `insert_only`, and `view_name` parameters.

This PR updates the serialisation method to use the correct serial name when serialising the schema.

We recently got some feedback that `@Serializable` should be avoided for public interfaces/classes as these can increase the objective C header size for the Swift SDK. This PR also removes `@Serializable` from the Schema definition classes. We still use `@Serializable` for other public classes, this is not addressed in this PR. In practice the changes here make the header file about 25 lines smaller.  

This PR also adds the ability to update the schema after a PowerSync client has been initialised.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
+++ b/core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
@@ -13,7 +13,7 @@ import kotlinx.coroutines.launch
 
 internal class ConnectionPool(
     factory: () -> PsSqlDriver,
-    size: Int = 5,
+    val size: Int = 5,
     private val scope: CoroutineScope,
 ) {
     private val available = Channel<Pair<TransactorDriver, CompletableDeferred<Unit>>>()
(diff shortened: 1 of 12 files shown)


core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
```diff
@@ -56,6 +56,28 @@ internal class ConnectionPool(
         }
     }
 
+    suspend fun <R> withAllConnections(action: suspend (connections: List<TransactorDriver>) -> R): R {
+        val obtainedConnections = mutableListOf<Pair<TransactorDriver, CompletableDeferred<Unit>>>()
+
+        try {
+            // Try and get all the connections
```
simolus3, line 63: Maybe expand this comment to mention that it gets all connections because we're holding on to all of them simultaneously, so there's no way we'd accidentally receive a connection twice.

(or maybe don't if you think that's obvious, it took me a short while to grasp that this is sound)

simolus3 (approved):
I'm happy with these changes 👍

simolus3 (approved):
Thanks!

core/src/commonMain/kotlin/com/powersync/db/internal/ConnectionPool.kt
stevensJourney: That's a fair point. I've expanded the comment :)

PR #155, opened 2025-03-27
Report download progress

simolus3:
This adds interfaces for tracking download progress during a sync operation. Following the design doc:

1. The `ProgressWithOperations` class provides information about how many rows have been downloaded already and how many need to be downloaded in total to complete the current sync (and uses that to provide a `fraction` getter).
2. The `SyncDownloadProgress` class keeps track of downloaded operations for each bucket. It can compute aggregate `ProgressWithOperations` results for the next complete sync (across all buckets) or up until a given priority (includes all buckets with that or a higher priority).
3. An instance of `SyncDownloadProgress` is made available from `SyncStatus`.

The implementation uses new columns added to `ps_buckets` tracking the operation count at the last checkpoint and the amount of downloaded operations since then:

1. When we receive a new checkpoint (or a diff), we fetch local progress counters from `ps_buckets` and use that to populate the initial progress status.
2. When receiving sync lines, the core extension will update the local counters. We update the `SyncDownloadProgress` instance in memory.
3. For a completed checkpoint, we reset the progress and update the "operation count at last checkpoint" column in the database.

I've also refactored some things around copying sync statutes internally (we don't have to write our own `copy` methods with special args to reset the upload + download error, optional arguments in Kotlin don't have to be constant so passing`downloadError = null` will actually reset the value).


compose/src/commonMain/kotlin/com/powersync/compose/DatabaseState.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.debounce
+import kotlin.time.Duration
+
+@OptIn(FlowPreview::class)
+@Composable
+public fun SyncStatus.composeState(debounce: Duration=Duration.ZERO): State<SyncStatusData> {
+    var flow: Flow<SyncStatusData> = asFlow()
+    if (debounce.isPositive()) {
```
stevensJourney, line 17: If my understanding is correct: a subtle "quirk" with the `debounce` operator is no values are emitted until a period of inactivity equal to the time window is experienced. Depending on the window size, this could result in no updates being emitted during periods of high sync volume/updates. 

IIRC we use debounce in the demos, but maybe we should use the `throttle` helper if we include this in the compose module?

simolus3: I agree that we shouldn't use `debounce`, but `throttle` is internal and I don't think exposing it for this is worth it.

I've removed the parameter to debounce the status now. While we can always add it back later without breaking anyone, I wouldn't expect the status to update often enough to justify a debouncing logic.

stevensJourney (approved):
I'm overall quite happy with the approach taken here. This is a very cool feature.

stevensJourney (approved):

stevensJourney (approved):

PR #156, opened 2025-03-27
Prepare static-sqlite-driver for publishing

simolus3:
It looks like we need to have some Kotlin sources for the package to get uploaded, otherwise we're missing some klibs. So, to fix the publishing step, this adds a bogus declaration.

I've also copied and adapted a `gradle.properties` file (not sure if that actually gets used though), enabled kotlinter and generally made some minor adjustments to the gradle file for that project to be closer to the others.


stevensJourney (approved):
LGTM 🙏

PR #157, opened 2025-03-27
Use structured concurrency for sync jobs

simolus3:
The jobs to download and upload data (as well as the one forwarding sync status changes) are all operating independently from each other, which can potentially cause issues stopping these jobs as once.

This adopts a `SupervisorJob` for the three, allowing us to cancel just one job in `disconnect` that's guaranteed to get them all.

I've also slightly refactored our test client so that it properly forwards cancellations (when we cancel the sync jobs, that propagates to `SyncStream.streamingSyncRequest` in the `request.execute` block). With the mock HTTP engine from ktor we unfortunately have no access to that scope, but by implementing a fairly trivial custom engine we can implement the same functionality while also reporting downstream listeners being cancelled.
This then allows testing that calling `disconnect()` or `close()` on a database does indeed close the underlying HTTP response stream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.0-BETA29 (unreleased)
+
+* Fix potential race condition between jobs in `connect()` and `disconnect()`.
+
 ## 1.0.0-BETA28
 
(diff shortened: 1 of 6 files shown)


stevensJourney (approved):
LGTM

stevensJourney (approved):

PR #158, opened 2025-04-02
Add User-Agent and remove User-Id

LucDeCaf:
Adds User-Agent in similar fashion to Dart and JS repos

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -9,6 +9,7 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.TestExecutable
 import org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest
 import org.jetbrains.kotlin.gradle.tasks.KotlinTest
 import org.jetbrains.kotlin.konan.target.Family
+import com.codingfeline.buildkonfig.compiler.FieldSpec.Type.STRING
 
 
 plugins {
@@ -21,6 +22,7 @@ plugins {
     id("com.powersync.plugins.sonatype")
     alias(libs.plugins.mokkery)
     alias(libs.plugins.kotlin.atomicfu)
+    alias(libs.plugins.buildKonfig)
 }
 
 val binariesFolder = project.layout.buildDirectory.dir("binaries/desktop")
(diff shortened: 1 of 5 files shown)


core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -449,6 +469,8 @@ internal class SyncStream(
     }
 }
 
+internal expect fun getOS(): String
```
simolus3, line 472: I'd prefer this to be in a separate file (e.g. `utils/OSInfo.kt`).

core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -172,7 +190,7 @@ internal class SyncStream(
                 contentType(ContentType.Application.Json)
                 headers {
                     append(HttpHeaders.Authorization, "Token ${credentials.token}")
-                    append("User-Id", credentials.userId ?: "")
+                    append("User-Agent", powerSyncUserAgent())
```
simolus3, line 193: Given that we're doing this on all requests, a better place to put this may be in `init`:

```Kotlin
fun HttpClientConfig<*>.configureClient() {
    install(HttpTimeout)
    install(ContentNegotiation)

    defaultRequest {
        headers {
            append("User-Agent", $LIBRARY_NAME/$LIBRARY_VERSION ${getOS()}")
        }
    }
}
```

simolus3 (commented):
> must check if can extract version string from `SDK_INT`

I think including `Build.VERSION` and `Build.RELEASE_OR_CODENAME` should be enough here.

core/build.gradle.kts
```diff
@@ -315,8 +317,19 @@ android {
...
+}
+
+buildkonfig {
+    packageName = "com.powersync.core"
+    defaultConfigs {
+        buildConfigField(STRING, "LIBRARY_VERSION", version.toString())
+
+        // TODO: Swift SDK relies on this too.
+        // Find out how to add a build flag to toggle between "powersync-kotlin" and "powersync-swift".
+        buildConfigField(STRING, "LIBRARY_NAME", "powersync-kotlin")
```
simolus3, line 332: I think we don't have to worry about this. We'll likely add an API for clients to add their own headers in the future. The Swift SDK could use that to always override the `User-Agent` header if necessary. So it should be fine if this always reports `powersync-kotlin` for now.

PR #159, opened 2025-04-03
Attachments package

stevensJourney:
# Overview

This adds attachments helper functionality to the PowerSync Core module.

For full context. See the [README](https://github.com/powersync-ja/powersync-kotlin/blob/attachments-package/core/src/commonMain/kotlin/com/powersync/attachments/README.md) 

This implementation is a combination of the approach taken in the JS and Dart implementations.
* The syncing strategy is adapted from JS. The Dart implementation requires developers to manually diff the watched attachments and local attachments state.
* The JS implementation is a single, large, abstract class. This implementation borrows from the Dart implementation and splits the components into an `AttachmentsService` which handles Attachment state persisting and a Syncing service which processes the queue. 
* The `AttachmentQueue` here is not abstract, it provides a common and simple default implementation for certain operations. The class is `open` and can be extended if developers require custom functionality. 

# Demo

See the small video demonstrating syncing attachments between two devices

https://github.com/user-attachments/assets/3b421448-7dda-452b-9d98-055734b92692

TODOS
- [x] Complete unit tests
- [x] Use in demos 
   - [x]  Added to Android Supabase Demo


core/src/commonMain/kotlin/com/powersync/attachments/Attachment.kt
```diff
@@ -0,0 +1,76 @@
...
+     */
+    val timestamp: Long = 0,
+    /**
+     * Attachment filename e.g. `[id].jpg`
+     */
+    val filename: String,
+    /**
+     * Current attachment state
+     */
+    val state: Int = AttachmentState.QUEUED_DOWNLOAD.ordinal,
```
simolus3, line 39: Not using the `AttachmentState` type here makes accessing the state from user code a bit awkward (unless I'm missing something they'd then have to look up the value through the index?)

So can we make this:

```suggestion
    val state: AttachmentState = AttachmentState.QUEUED_DOWNLOAD,
```

And do something like `state = AttachmentState.entries[cursor.getLong("state").toInt()]` when mapping from rows?

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+    }
+}
+
+/**
+ * Class used to implement the attachment queue
+ * Requires a PowerSyncDatabase, an implementation of
+ * AbstractRemoteStorageAdapter and an attachment directory name which will
+ * determine which folder attachments are stored into.
+ */
+public open class AttachmentQueue(
```
simolus3, line 59: ~~Does this have to be `open`? We're not extending it at the moment, and making classes `open` in the future is always easier than the other way around :)~~ Nevermind, saw the logic to create filenames that should be replacable.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+     */
+    private val archivedCacheLimit: Long = 100,
+    /**
+     * Throttles remote sync operations triggering
+     */
+    private val syncThrottleDuration: Duration = 1.seconds,
+    /**
+     * Creates a list of subdirectories in the {attachmentDirectoryName} directory
+     */
+    private val subdirectories: List<String>? = null,
```
simolus3, line 125: I remember a FlutterFlow user asking for dynamic subdirectories (because they have something like a `projects` table and want to store images in `$project/$image.png`. Users can be added and removed from projects dynamically).

So I'm just wondering if it might be easier to just support slashes as path separators (creating the directory automatically) instead of listing subdirectories statically. But consistency with our other SDKs is also important, so I'm not sure if it's worth it.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+    @Throws(PowerSyncException::class, CancellationException::class)
+    public suspend fun stopSyncing(): Unit =
+        runWrappedSuspending {
+            mutex.withLock {
+                if (closed) {
+                    return@runWrappedSuspending
+                }
+
+                syncStatusJob?.cancel()
+                syncStatusJob?.join()
```
simolus3, line 239: Might need an additional import

```suggestion
                syncStatusJob?.cancelAndJoin()
```

(same thing in `close()`

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+                    throw Exception("Attachment queue has been closed")
+                }
+                // Ensure the directory where attachments are downloaded, exists
+                localStorage.makeDir(attachmentsDirectory)
+
+                subdirectories?.forEach { subdirectory ->
+                    localStorage.makeDir(Path(attachmentsDirectory, subdirectory).toString())
+                }
+
+                val scope = CoroutineScope(Dispatchers.IO)
```
simolus3, line 195: This effectively launches the `syncStatusJob` inside the `GlobalScope` without structured concurrency. I think this should take a `CoroutineScope` from which it inherits. We could:

1. Make `startSync()` an extension function on `CoroutineScope` (`public suspend fun CoroutineScope.startSync()`), a pattern I've repeatedly seen in Kotlin.
2. Make users pass a `CoroutineScope` to the`AttachmentQueue` (it can be the `GlobalScope` by default, but it should at least be possible to override this).
3. Make the coroutine scope for the database implementation `internal` and re-use that here, since we have access to the database.

Then regardless of where we get the outer scope from, we can use `val scope = CoroutineScope(outerScope.newCoroutineContext(Dispatchers.IO))` here.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+     *
+     * This method can be overriden for custom behaviour.
+     */
+    @Throws(PowerSyncException::class, CancellationException::class)
+    public open suspend fun saveFile(
+        data: Flow<ByteArray>,
+        mediaType: String,
+        fileExtension: String? = null,
+        metaData: String? = null,
+        updateHook: (context: ConnectionContext, attachment: Attachment) -> Unit,
```
simolus3, line 392: I wonder if this should be suspending just in case

core/src/commonMain/kotlin/com/powersync/attachments/storage/IOLocalStorageAdapter.kt
```diff
@@ -0,0 +1,114 @@
...
+import kotlinx.coroutines.withContext
+import kotlinx.io.Buffer
+import kotlinx.io.buffered
+import kotlinx.io.files.Path
+import kotlinx.io.files.SystemFileSystem
+
+/**
+ * Storage adapter for local storage using the KotlinX IO library
+ */
+public class IOLocalStorageAdapter : LocalStorage {
```
simolus3, line 21: I suppose for testing or if users want a custom file system, adding a `private val fileSystem: FileSystem = SystemFileSystem` and using that in the methods may be convenient?

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,278 @@
...
+public class SyncingService(
+    private val remoteStorage: RemoteStorage,
+    private val localStorage: LocalStorage,
+    private val attachmentsService: AttachmentService,
+    private val getLocalUri: suspend (String) -> String,
+    private val errorHandler: SyncErrorHandler?,
+    private val logger: Logger,
+    private val syncThrottle: Duration = 5.seconds,
+) {
+    private val scope = CoroutineScope(Dispatchers.IO)
```
simolus3, line 42: The parent scope should also be configurable here.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,278 @@
...
+     */
+    private val syncTriggerFlow = MutableSharedFlow<Unit>(replay = 0)
+
+    /**
+     * Starts syncing operations
+     */
+    public suspend fun startSync(period: Duration = 30.seconds): Unit =
+        mutex.withLock {
+            syncJob?.cancel()
+            syncJob?.join()
```
simolus3, line 57:
```suggestion
            syncJob?.cancelAndJoin()
```

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,278 @@
...
+                                // Triggers the sync process whenever an underlaying change to the
+                                // attachments table happens
+                                attachmentsService
+                                    .watchActiveAttachments(),
+                            )
+                                // We only use these flows to trigger the process. We can skip multiple invocations
+                                // while we are processing. We will always process on the trailing edge.
+                                // This buffer operation should automatically be applied to all merged sources.
+                                .buffer(1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
+                                .throttle(syncThrottle.inWholeMilliseconds)
```
simolus3, line 75: Given that it's internal, I wonder if we should just change `throttle` to take a `Duration` 🤔

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,278 @@
...
+                                        }
+                                    }
+                                }
+                        }
+
+                    val periodicJob =
+                        launch {
+                            logger.i("Periodically syncing attachments")
+                            syncTriggerFlow.emit(Unit)
+                            delay(period)
```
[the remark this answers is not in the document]
simolus3: The bot sure seems to have a point

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,278 @@
...
+                        }
+
+                    watchJob.join()
+                    periodicJob.join()
```
simolus3, line 109: Mainly out of my own interest, but are these necessary? It was my understanding that the parent job would be active as long as it has an active child job too.

demos/android-supabase-todolist/app/src/main/java/com/powersync/androidexample/App.kt
```diff
@@ -26,25 +31,57 @@ import com.powersync.demos.screens.SignUpScreen
...
+        remember {
+            SupabaseConnector(
+                powerSyncEndpoint = BuildConfig.POWERSYNC_URL,
+                supabaseUrl = BuildConfig.SUPABASE_URL,
+                supabaseKey = BuildConfig.SUPABASE_ANON_KEY,
+                storageBucket = BuildConfig.SUPABASE_ATTACHMENT_BUCKET,
+            )
+        }
+
+    val db = remember { PowerSyncDatabase(driverFactory, schema, dbFilename = "333.sqlite") }
```
simolus3, line 50: Is the `dbFilename` a leftover from debugging?

simolus3 (commented):
I'm very happy with the approach overall (also thanks for the in-depth documentation in the readme). I didn't try the example yet, but I have some comments (most of which are very minor though).

demos/android-supabase-todolist/local.properties.example
```diff
@@ -11,6 +11,7 @@ sdk.dir=/Users/dominic/Library/Android/sdk
 
 SUPABASE_URL=https://foo.supabase.co
 SUPABASE_ANON_KEY=foo
+SUPABASE_ATTACHMENT_BUCKET=media # optional attachment bucket
```
benitav, line 14: I would rename to SUPABASE_STORAGE_BUCKET since that's what it's called in Supabase

core/src/commonMain/kotlin/com/powersync/attachments/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Alpha Release
+
+Attachment helpers are currently in an alpha state, intended strictly for testing. Expect breaking
+changes
+and instability as development continues.
+
+Do not rely on this package for production use.
+
+## Usage
```
benitav, line 15: Can we add a link to the demo app with the example implementation. Maybe under an ## Example subheading

core/src/commonMain/kotlin/com/powersync/attachments/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+// Pass the handler to the queue constructor
+val queue = AttachmentQueue(
+//    ...,
+    errorHandler = errorHandler,
+//    ...
+)
+```
+
+# Implementation details
```
benitav, line 213: Think this should be a ## level heading - and update subsequent ones accordingly?

core/src/commonMain/kotlin/com/powersync/attachments/README.md
```diff
@@ -0,0 +1,320 @@
...
+* Deletes the attachment on the local filesystem
+* Updates the record to the `QUEUED_DELETE` state
+* Allows removing assignments to relational data.
+    * It's important to unassign the attachment from relational data since this data is constantly
+      watched and should always represent the attachment queue state. Failure to unassign the
+      attachment could result in a failed delete.
+
+### Expire Cache
+
+When PowerSync removes a record, as a result of coming back online or conflict resolution for
```
benitav, line 310: Can you help me understand this example scenario in different words please? I'm not sure what kind of record we're referring to here, or why PowerSync is removing stuff when coming back online. (I realize we copied this from RN)

core/src/commonMain/kotlin/com/powersync/attachments/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Key Assumptions
+
+- Each attachment should be identifiable by a unique ID.
+- Attachments are immutable.
+- Relational data should contain a foreign key column that references the attachment ID.
+- Relational data should reflect the holistic state of attachments at any given time. An existing
+  local attachment will deleted locally if no relational data references it.
+
+### Example
```
benitav, line 28: I assume ### Steps to implement below also form part of this example? The headings being on the same level makes it confusing.

core/src/commonMain/kotlin/com/powersync/attachments/README.md
```diff
@@ -0,0 +1,320 @@
...
+val schema = Schema(
+    UserRow.table,
+    // Includes the table which stores attachment states
+    createAttachmentsTable("attachments")
+)
+```
+
+The `createAttachmentsTable` function defines the local only attachment state storage table.
+
+An attachments table definition can be created with the following options.
```
benitav, line 55: I find it a bit confusing to have the list of options defined here under "example". Seems like something that should rather be listed under "Implementation details" since it's true for all implementations? Then just reference it here.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
stevensJourney: This hook currently is executed inside the write transaction which is unfortunately not a suspending context.

demos/android-supabase-todolist/app/src/main/java/com/powersync/androidexample/SupabaseRemoteStorage.kt
```diff
@@ -0,0 +1,33 @@
...
+) : RemoteStorage {
+    override suspend fun uploadFile(
+        fileData: Flow<ByteArray>,
+        attachment: Attachment,
+    ) {
+        // Supabase wants a single ByteArray
+        val buffer = ByteArray(attachment.size!!.toInt())
+        var position = 0
+        fileData.collect {
+            System.arraycopy(it, 0, buffer, position, it.size)
```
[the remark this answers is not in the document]
stevensJourney: Just for context: The only reason the attachment demo is in the Android project is due to not prioritising a Camera/media service implementation for all the different platforms in the compose demo. IMO this can be scheduled for later if necessary.

demos/android-supabase-todolist/app/src/main/java/com/powersync/androidexample/SupabaseRemoteStorage.kt
```diff
@@ -0,0 +1,33 @@
+package com.powersync.androidexample
+
+import com.powersync.attachments.Attachment
+import com.powersync.attachments.RemoteStorage
+import com.powersync.connector.supabase.SupabaseConnector
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+
+class SupabaseRemoteStorage(
```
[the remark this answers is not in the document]
stevensJourney: Fair point. I've moved this to the Supabase connector project.

core/src/commonMain/kotlin/com/powersync/attachments/Attachment.kt
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Enum for the attachment state
+ */
+public enum class AttachmentState {
+    QUEUED_DOWNLOAD,
+    QUEUED_UPLOAD,
+    QUEUED_DELETE,
+    SYNCED,
+    ARCHIVED,
```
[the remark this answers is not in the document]
stevensJourney: 👍 Added comments here

core/src/commonMain/kotlin/com/powersync/attachments/Attachment.kt
stevensJourney: Yeah that is nicer. I've implemented this.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
stevensJourney: Noice. Implemented this.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
```diff
@@ -0,0 +1,489 @@
...
+    @Throws(PowerSyncException::class, CancellationException::class)
+    public open suspend fun deleteFile(
+        attachmentId: String,
+        updateHook: (context: ConnectionContext, attachment: Attachment) -> Unit,
+    ): Attachment =
+        runWrappedSuspending {
+            attachmentsService.withLock { attachmentContext ->
+                val attachment =
+                    attachmentContext.getAttachment(attachmentId)
+                        ?: throw Exception("Attachment record with id $attachmentId was not found.")
```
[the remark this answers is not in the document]
stevensJourney: Updated to use `error` now.

core/src/commonMain/kotlin/com/powersync/attachments/storage/IOLocalStorageAdapter.kt
stevensJourney: That's a fair recommendation. I've implemented this.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
stevensJourney: This is configurable now

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
stevensJourney: True. This was not necessary. Removed it.

demos/android-supabase-todolist/app/src/main/java/com/powersync/androidexample/App.kt
stevensJourney: Jip. Good catch. Removed this.

demos/android-supabase-todolist/local.properties.example
stevensJourney: 👍 Changed this.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
stevensJourney: Yeah. Changed this to a Duration now.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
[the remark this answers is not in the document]
stevensJourney: A good bot indeed. I've implemented this.

core/src/commonMain/kotlin/com/powersync/attachments/AttachmentQueue.kt
stevensJourney: I suppose a dev could override the local storage implementation to upsert directories if they required that functionality.

demos/android-supabase-todolist/local.properties.example
```diff
@@ -11,6 +11,7 @@ sdk.dir=/Users/dominic/Library/Android/sdk
 
 SUPABASE_URL=https://foo.supabase.co
 SUPABASE_ANON_KEY=foo
+SUPABASE_STORAGE_BUCKET=media # optional attachment bucket, set to null to disable
```
simolus3, line 14: IIRC our guides don't mention this bucket, so I wonder if we should set it to null in the template and then make this an opt-in feature? That's the way things are on our other SDKs, and it ensures running the example works even without the bucket being set up.

core/src/commonMain/kotlin/com/powersync/attachments/storage/IOLocalStorageAdapter.kt
```diff
@@ -0,0 +1,117 @@
...
+import kotlinx.io.files.FileSystem
+import kotlinx.io.files.Path
+import kotlinx.io.files.SystemFileSystem
+import kotlinx.io.readByteArray
+
+/**
+ * Storage adapter for local storage using the KotlinX IO library
+ */
+public class IOLocalStorageAdapter : LocalStorage {
+    private val fileSystem: FileSystem = SystemFileSystem
```
simolus3, line 23: I think we can now make this configurable too:

```suggestion
public class IOLocalStorageAdapter(
    private val fileSystem: FileSystem = SystemFileSystem
) : LocalStorage {
```

core/src/commonMain/kotlin/com/powersync/attachments/implementation/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.powersync.db.getString
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+/**
+ * Service for interacting with the local attachment records.
+ */
+public open class AttachmentServiceImpl(
```
simolus3, line 17: Can this be `internal`?

core/src/commonMain/kotlin/com/powersync/attachments/implementation/AttachmentContextImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+import com.powersync.attachments.Attachment
+import com.powersync.attachments.AttachmentContext
+import com.powersync.attachments.AttachmentState
+import com.powersync.db.getString
+import com.powersync.db.internal.ConnectionContext
+import kotlinx.datetime.Clock
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+public open class AttachmentContextImpl(
```
simolus3, line 14: Same here, wondering if this can be `internal`?

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,311 @@
...
+ * @property localStorage The local storage implementation for managing files locally.
+ * @property attachmentsService The service for managing attachment states and operations.
+ * @property getLocalUri A function to resolve the local URI for a given filename.
+ * @property errorHandler Optional error handler for managing sync-related errors.
+ * @property logger Logger instance for logging sync operations and errors.
+ * @property syncThrottle The minimum duration between consecutive sync operations.
+ * @property scope The coroutine scope used for managing sync operations.
+ */
+public class SyncingService
+    @OptIn(DelicateCoroutinesApi::class)
```
simolus3, line 47: IntelliJ says we don't need this annotation here :)

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
```diff
@@ -0,0 +1,311 @@
...
+ * @property remoteStorage The remote storage implementation for handling file operations.
+ * @property localStorage The local storage implementation for managing files locally.
+ * @property attachmentsService The service for managing attachment states and operations.
+ * @property getLocalUri A function to resolve the local URI for a given filename.
+ * @property errorHandler Optional error handler for managing sync-related errors.
+ * @property logger Logger instance for logging sync operations and errors.
+ * @property syncThrottle The minimum duration between consecutive sync operations.
+ * @property scope The coroutine scope used for managing sync operations.
+ */
+public class SyncingService
```
simolus3, line 46: Also here, given that this appears to be set up by the `AttachmentQueue` only, can it be `internal`?

simolus3 (commented):
I'm happy with these changes, the remaining comments are very minor items.

demos/android-supabase-todolist/local.properties.example
stevensJourney: Good point. I've updated the example default to `null`

core/src/commonMain/kotlin/com/powersync/attachments/storage/IOLocalStorageAdapter.kt
stevensJourney: 👍 Implemented

core/src/commonMain/kotlin/com/powersync/attachments/implementation/AttachmentContextImpl.kt
stevensJourney: Other SDKs usually allow overriding the AttachmentQueue implementation. I'd like to keep these defaults public in order for devs to extend them if necessary. I've added a comment to these classes.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
stevensJourney: This was left from experimenting with defaulting the coroutine context to `GlobalScope`. Good catch, I've removed this now.

core/src/commonMain/kotlin/com/powersync/attachments/sync/SyncingService.kt
stevensJourney: This was left from experimenting with defaulting the coroutine context to `GlobalScope`. Good catch, I've removed this now.

simolus3 (approved):

PR #160, opened 2025-04-03
Fix checkpoints during uploads not being applied

simolus3:
This is the Kotlin port of https://github.com/powersync-ja/powersync-js/pull/558, see that PR for a more complete description.

I also had to fix some other quirks in the sync implementation here, notably:

1. The `triggerCrudUpload()` call when receiving keep-alive messages used to block the entire sync stream from receiving subsequent lines until the upload was completed. This changes that to spawn an async job that is not awaited (the trigger listening on `ps_crud` will still await this though).
2. There was no way to abort sync iterations (used e.g. for checkpoints with a checksum mismatch). This adds that functionality.


stevensJourney (dismissed):
LGTM

stevensJourney (approved):

Issue #161, opened 2025-04-05
`AccessDeniedException` when deleting temporary DLL file on Windows AppData folder (JVM)


luanrodrigsr:
Hi!

After investigating the log of my JVM application, i notice an error when the user close it on Windows. 

**SDK version:** 1.0.0-BETA28
**Platform OS and version:** Windows 11 (JVM) amd64

```Logcat
DEBUG: Database SQLiteVersion: 3.49.1
DEBUG: Database PowerSyncVersion: 0.3.12/0386fb5d
ERROR: [DATABASE] O usuario nao esta conectado.
Exception in thread "Thread-4" java.nio.file.AccessDeniedException: C:\Users\USER\AppData\Local\Temp\powersync4601466118483182314.dll
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:89)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:273)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:109)
	at java.base/java.nio.file.Files.deleteIfExists(Files.java:1191)
	at com.powersync.ExtractLibKt.extractLib$lambda$0(ExtractLib.kt:30)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

![Image](https://github.com/user-attachments/assets/e094aa79-01c9-46c3-990a-8907d2ebe8dd)

I'm not sure about the impact and severity of this exception right now, but I decided to share it here. 

It seems as a failed attempt to delete a temporary DLL file due to insufficient permission.

```Kotlin

val tmpPath = createTempFile("$prefix$fileName", ".$extension")
Runtime.getRuntime().addShutdownHook(Thread { tmpPath.deleteIfExists() })

```

stevensJourney:
Thanks for reporting this! I was able to reproduce this by running our unit tests on a Windows machine. Strangely our Windows CI runner does not exhibit the same behaviour.  
For context. We are currently copying our PowerSync SQLite extension DLL to a temporary file, which is scheduled for delete when the process shuts down. I believe we should be able to improve this workflow. 
A fix for this should be released soon, most likely next week.

[...]

PR #162, opened 2025-04-07
[JVM Windows] Process shutdown bug

stevensJourney:
# Overview

For JVM we currently find the PowerSync SQLite core resource's path then copy it to a temporary file. The temporary file is used when loading the extension in SQLite. The temporary file is deleted on process shutdown. 

https://github.com/powersync-ja/powersync-kotlin/issues/161 raised that on Windows an issue occurs when deleting the temporary file.

It looks like this method, of using a temporary file, was used in order to resolve the extension's path in a format compatible with Windows. Typically the `R::class.java.getResource(path)` result is of the format `/c:/...` which does not seem to work as an extension path in SQLite.

The logic has been updated to avoid the use of temporary files altogether. The binary resource file path is now used to load the SQLite extension.

### DIFF
--- a/core/src/jvmMain/kotlin/com/powersync/ExtractLib.kt
+++ b/core/src/jvmMain/kotlin/com/powersync/ExtractLib.kt
@@ -30,6 +30,5 @@ internal fun extractLib(fileName: String): String {
      * Wrapping in a File handle resolves the URI to a path usable by SQLite.
      * This is particularly relevant on Windows.
      */
-    val result = File(resourceURI.path).path.toString()
-    return result
+    return File(resourceURI.path).path.toString()
 }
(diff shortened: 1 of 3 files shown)


core/src/jvmMain/kotlin/com/powersync/ExtractLib.kt
```diff
@@ -26,14 +23,12 @@ internal fun extractLib(fileName: String): Path {
...
-        tmpPath.outputStream().use { output ->
-            input.copyTo(output)
-        }
-    }
-
-    return tmpPath
+    /**
+     * Wrapping in a File handle resolves the URI to a path usable by SQLite.
+     * This is particularly relevant on Windows.
+     */
```
simolus3, line 32: > Wrapping in a File handle resolves the URI to a path usable by SQLite.

For my own understanding, is this to transform the path into the native representation (because `resourceURI.path` will always be a URL-style path which is not supported on Windows?). Maybe the comment could be a little clearer on why it's relevant.

Unrelated nit: I prefer `//` comments inside of methods, this one isn't part of any KDoc declaration either way.

simolus3 (approved):
Thanks 👍

simolus3:
Let's see if merging the test changes does something about the flakiness 🤞

PR #163, opened 2025-04-07
WIP API Reference 

stevensJourney:
# Overview

This adds an automatically generated HTML API Reference using [Dokka](https://github.com/Kotlin/dokka).

This can be served by running `./gradlew serveDokka` in the repo root.

image

# TODOS

- [ ] General comment cleanup
- [ ] Add entries for Connectors module
- [ ] Github pages publish flow

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,9 @@
+import com.sun.net.httpserver.HttpExchange
+import com.sun.net.httpserver.HttpServer
+import java.net.InetSocketAddress
+import java.net.URLDecoder
+import java.nio.file.Files
+
 plugins {
     alias(libs.plugins.jetbrainsCompose) apply false
     alias(libs.plugins.compose.compiler) apply false
@@ -16,6 +22,7 @@ plugins {
     alias(libs.plugins.kotlinter) apply false
     alias(libs.plugins.keeper) apply false
     alias(libs.plugins.kotlin.atomicfu) apply false
+    id("org.jetbrains.dokka") version "2.0.0"
 }
 
 allprojects {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -54,6 +62,56 @@ subprojects {
...
 
-tasks.register<Delete>("clean") {
+tasks.getByName<Delete>("clean") {
     delete(rootProject.layout.buildDirectory)
 }
+
+// Merges individual module docs into a single HTML output
+dependencies {
+    dokka(project(":core:"))
+    dokka(project(":connectors:supabase"))
```
simolus3, line 72: I assume this should also include `:compose`?

build.gradle.kts
```diff
@@ -54,6 +62,56 @@ subprojects {
...
+}
+
+dokka {
+    moduleName.set("PowerSync Kotlin")
+}
+
+// Serve the generated Dokka documentation using a simple HTTP server
+// File changes are not watched here
+tasks.register("serveDokka") {
+    dependsOn("dokkaGenerate")
```
simolus3, line 82: Let's add a `group = "dokka"` line here so that it shows up next to the other tasks in IDEs:

Task overview

simolus3 (commented):
This looks like a really nice thing to have 👍 

The images don't load for me locally, not sure if that's a setup issue or if it would also happen when deploying to GH pages:

grafik

I also noticed that the `BuildConfig` class now gets kind of ugly because it's not in a package. Do you know if that class is used across packages? I'm in favor of making it `internal` to hide the `[root]` package on the docs.

build.gradle.kts
stevensJourney: Good point! I forgot that module existed 😁 . Added it now.

simolus3 (dismissed):
LGTM 👍

build.gradle.kts
stevensJourney: Added a group.

simolus3 (approved):

PR #164, opened 2025-04-07
[Fix] IOS Migrations

stevensJourney:
# Overview

We currently create multiple readonly connections via the iOS driver. The iOS driver attempts to run schema migrations whenever a driver is created. It looks like these migrations are not synchronized internally. This causes errors of the form

```
Sqlite operation failure database is locked attempted to run migration and failed. closing connection
```

See example failed [test](https://github.com/powersync-ja/powersync-kotlin/actions/runs/14302972547/job/40081077924?pr=162)

This changeset avoids attempting to run migrations on read only drivers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## 1.0.0-BETA29 (unreleased)
 
 * Fix potential race condition between jobs in `connect()` and `disconnect()`.
+* [iOS] Fixed issue where automatic driver migrations would fail with the error:
+```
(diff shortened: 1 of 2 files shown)


stevensJourney:
> Nice 👍 I wonder if we should just always pass `NO_VERSION_CHECK` given that we don't actually apply any schema. But it's good to have the fix in first, we can check refactoring this and removing `InternalSchema` afterwards.

Agreed. We probably could and should refactor a lot of the current SQLDelight iOS (and potentially others) driver at some point.

PR #165, opened 2025-04-07
Refactor tests for better isolation

simolus3:
This refactors our tests opening databases (`DatabaseTest` and `SyncIntegrationTest`) to:

1. No longer use `GlobalScope` when opening databases. Instead, we use the test scope we receive from `runTest`. Thanks to structured concurrency, this ensures that no async database code can "leak" out of the test.
2. Generate a random file name for each test, ensuring that two different tests won't interfere with each other.
3. In some places, start using the assertions library from [Kotest](https://kotest.io/docs/assertions/assertions.html) instead of `kotlin-test`. These assertions are more expressive and lead to much better error messages (e.g. `list shouldHaveSize 3` will actually print contents of the list on violations, `assertEquals(3, list.size)` is much less helpful).

The first step required some other changes, in particular:

1. We can no longer use `@Setup` in tests to open databases, because that function is non-suspending and doesn't run in the test's coroutine scope. Instead, I wrote a utility (the `databaseTest` function) that creates an `ActiveDatabaseTest` instance encapsulating all that logic. This ensures that the full lifecycle (initializing databases, running tests, closing databases) happens within the test's coroutine scope.
2. The `runBlocking` initializer in `PowerSyncDatabaseImpl` is incompatible with the new tests. I think this is because the connection pool also uses the coroutine scope from tests now (which is no longer a `GlobalScope`). And the `runBlocking` then prevents that scope from progressing? I'm not 100% sure about this, but there was some deadlock with `runBlocking` when opening the database.
A workaround that I think is also beneficial for apps is to move the initialization into a `Job` that is then awaited before each other method. This is the same pattern we use for Dart and JavaScript as well, and ensures we're not accidentally blocking any UI thread that might be opening the PowerSync database.

### DIFF
--- a/core/src/commonIntegrationTest/kotlin/com/powersync/SyncIntegrationTest.kt
+++ b/core/src/commonIntegrationTest/kotlin/com/powersync/SyncIntegrationTest.kt
@@ -15,7 +15,6 @@ import com.powersync.testutils.databaseTest
 import com.powersync.testutils.waitFor
 import com.powersync.utils.JsonUtil
 import dev.mokkery.verify
-import io.kotest.assertions.withClue
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
 import kotlinx.coroutines.DelicateCoroutinesApi
@@ -48,10 +47,8 @@ class SyncIntegrationTest {
                 turbine.cancel()
             }
 
-            // Closing the database should have closed the channel
-            withClue("Should have closed sync stream") {
-                syncLines.isClosedForSend shouldBe true
-            }
+            // Closing the database should have closed the channel.
+            waitFor { syncLines.isClosedForSend shouldBe true }
         }
 
     @Test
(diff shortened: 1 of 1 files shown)


core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -98,48 +100,56 @@ internal class PowerSyncDatabaseImpl(
...
-            logger.d { "SQLiteVersion: $sqliteVersion" }
-            powerSyncVersion =
-                internalDb.get("SELECT powersync_rs_version()") { it.getString(0)!! }
-            checkVersion(powerSyncVersion)
-            logger.d { "PowerSyncVersion: ${getPowerSyncVersion()}" }
-
-            internalDb.writeTransaction { tx ->
-                tx.getOptional("SELECT powersync_init()") {}
-            }
+    private suspend fun initialize() {
```
stevensJourney, line 107: Thanks for improving this!

stevensJourney (approved):
Excellent work!

stevensJourney (approved):

PR #173, opened 2025-04-22
Set temp_store_directory on Android

simolus3:
SQLite tries to write some temporary data, such as large intermediate materialized views or indexes, to the file system. For this, it attempts to use paths in `/tmp` or the current working directory. Due to sandboxing, both directories are unavailable on Android though. This can cause [errors that users were running into](https://discord.com/channels/1138230179878154300/1357681026939486369/1362417181802893482).

As a workaround, this overwrites the temporary directory to point to the cache directory from the app's `Context`.

### DIFF
--- a/core-tests-android/src/androidTest/java/com/powersync/AndroidDatabaseTest.kt
+++ b/core-tests-android/src/androidTest/java/com/powersync/AndroidDatabaseTest.kt
@@ -227,5 +227,8 @@ class AndroidDatabaseTest {
     fun canCreateTemporaryTable() = runTest {
         database.execute("PRAGMA temp_store = 1;") // Store temporary data as files
         database.execute("CREATE TEMP TABLE my_tbl (content ANY) STRICT;")
+        for (i in 0..128) {
+            database.execute("INSERT INTO my_tbl VALUES (randomblob(1024 * 1024))")
+        }
     }
 }
--- a/core-tests-android/src/androidTest/java/com/powersync/AndroidDatabaseTest.kt
+++ b/core-tests-android/src/androidTest/java/com/powersync/AndroidDatabaseTest.kt
@@ -222,4 +222,10 @@ class AndroidDatabaseTest {
             // The exception messages differ slightly between drivers
             assertEquals(exception.message!!.contains("write a readonly database"), true)
         }
+
+    @Test
+    fun canCreateTemporaryTable() = runTest {
+        database.execute("PRAGMA temp_store = 1;") // Store temporary data as files
+        database.execute("CREATE TEMP TABLE my_tbl (content ANY) STRICT;")
+    }
 }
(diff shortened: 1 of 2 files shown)


core-tests-android/src/androidTest/java/com/powersync/AndroidDatabaseTest.kt
```diff
@@ -222,4 +222,10 @@ class AndroidDatabaseTest {
             // The exception messages differ slightly between drivers
             assertEquals(exception.message!!.contains("write a readonly database"), true)
         }
+
+    @Test
+    fun canCreateTemporaryTable() = runTest {
+        database.execute("PRAGMA temp_store = 1;") // Store temporary data as files
+        database.execute("CREATE TEMP TABLE my_tbl (content ANY) STRICT;")
```
simolus3, line 229: It's worth noting that this test passes even without the pragma, but I'm not sure why. My understanding is that SQLite should use temporary files here, but it apparently doesn't.

rkistner: I believe SQLite always keeps it in memory first, until it reaches a certain size threshold. I can't find any docs confirming this though.

simolus3: I tried inserting a 1MiB `randomblob` 128 times, but it still seems to pass without setting the temp directory.

rkistner: Another thing to check is the Android version it is being tested on - I remember this potentially only affecting newer Android versions. Although I can't find where the test emulator is specified here?

simolus3: I've just tested this manually with an SDK 35 emulator. An SDK-specific issue would surprise me though, we ship our own SQLite library along with the application. Or are you saying that some Android versions give apps their own `/tmp` that would be writable?

rkistner: Ok, I think I'm remembering anecdotes of temporary storage issues only being present on some Android versions, but I don't think we ever confirmed that.

Either way it doesn't quite explain how these tests worked, unless the emulator uses different filesystem permission structures from actual devices?

simolus3: > unless the emulator uses different filesystem permission structures from actual devices?

That sounds unlikely, the emulator is a VM running a real Android system.

> I think I'm remembering anecdotes of temporary storage issues only being present on some Android versions

I've asked the user reporting this to share the Android version they're using. Maybe it's not a problem on newer Android versions and that's why we didn't catch this.

Issue #167, opened 2025-04-13
UPDATE in transaction is not applied sometimes


Radiokot:
Hi guys, I have this weird issue with 1.0.0-BETA28, and I would greatly appreciate any help in debugging it.

Scenario:
- An expense logging screen of an Android app is opened via a desktop shortcut
- User selects account and enters the amount
- A `writeTransaction` is executed in `viewModelScope` which contains the following queries:
  - INSERT new row into the transfers table
  - SELECT current account balance from the accounts table (stored as TEXT and mapped into BigInteger)
  - UPDATE the accounts table, set corresponding account balance to current - transfer amount
- Once `writeTransaction` is completed successfully, the screen closes and so the app goes to background

The issue: sometimes the balance UPDATE is not applied, although the new row gets inserted into the transfers table within the same transaction.

On the other hand, if the same expense logging is done from the main app screen, so the app doesn't go to background on transaction success, the UPDATE is always applied, and the user immediately sees the updated value in the UI (via `database.watch()`)

Here is the log:

[log.txt](https://github.com/user-attachments/files/19726648/log.txt)

Radiokot:
I do not control this limit, I connect my `PowerSyncDatabase` with `SupabaseConnector`, pretty standard 🤔

simolus3:
Oh right, sorry 🤦 And that one actually uses the default limit, so I need to check where that `LIMIT 1` is coming from.

Radiokot:
![Image](https://github.com/user-attachments/assets/816b92c6-0bf3-4563-aa63-a5b0564eacd6)

Radiokot:
Hi @simolus3, I have new findings to share:
1. If I swap transfers INSERT and accounts UPDATE, so the update is executed first, then the update result is always preserved, and the transfer gets inserted locally as well!
2. However, when it immediately tries to sync the changes to Supabase when the app is already in background, the transfers INSERT (POST) **sometimes** is rejected because of RLS policy violation. Why? Because for this operation for some reason an anon role JWT is used, despite my `PowerSyncDatabase` gets connected with `SupabaseConnector` upon creation. When I then open the app later and the sync restarts, the INSERT eventually gets synced.

So, weird stuff.

Here you can see attempts to sync transfers INSERT with anon role JWT:
```
[...]

PR #175, opened 2025-04-23
Do not split transactions when uploading data

Radiokot:
Fix splitting DB transactions into separate CRUD transactions when uploading data.

Related to https://github.com/powersync-ja/powersync-kotlin/issues/167#issuecomment-2823551791


Radiokot:
Please, check if it builds, as I wasn't able to initialize Gradle for this project in IDEA.
@stevensJourney @simolus3 @kobiebotha

simolus3:
> Please, check if it builds, as I wasn't able to initialize Gradle for this project in IDEA.

I'll check, but could you post the error message if you have gotten any? Also, which OS are you on? We definitely want to make sure everyone can contribute here, so this is something we need to fix.

Radiokot:
Thank you. I finally was able to build it in IDEA, had to set Java version to 17 (picked 1.8 by default). The error was missing "org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_8" constant in one of the Gradle files.
Now it builds, and I fixed my commit (had syntax error in it).

PR #176, opened 2025-04-24
Add onChange API

stevensJourney:
# Overview

This adds an `onChange` method to the PowerSync client API.

This method watches a Set of table names and emits the subset of tables which changed whenever a relevant table has a changed.

The logic here is consistent with other SDKs. The input table names are flat mapped  to include watching changes for internal table names. The result is then converted to friendly table names.


core/src/commonMain/kotlin/com/powersync/db/Queries.kt
```diff
@@ -59,65 +84,91 @@ public interface Queries {
     ): RowType?
 
     /**
-     * Execute a read-only (SELECT) query every time the source tables are modified and return the results as a List in [Flow].
+     * Returns a [Flow] that emits whenever the source tables are modified.
+     *
+     * @param tables The set of tables to monitor for changes.
+     * @param throttleMs The minimum interval, in milliseconds, between queries. Defaults to null.
```
simolus3, line 90: This doesn't actually run a query.

I think it would also be good to mention what happens if the database updates more frequently than the `throttleMs`. IIRC we collect writes into a growing set that is them emitted later?

"Defaults to null" is also misleading since the implementation does apply a default (`DEFAULT_WATCH_THROTTLE`) internally when no value is given. From that sentence, I would assume that no throttling takes place when the default value of `null` is passed.

simolus3 (commented):
I'm happy with these changes, also thanks for updating the docs! I think `throttleMs` deserves a bit more detail in the docs, but it wasn't documented at all before so this is alreay a big upgrade 😄

core/src/commonMain/kotlin/com/powersync/db/Queries.kt
stevensJourney: Good point! This implementation actually wasn't accumulating the set of changed tables between throttles - an event would have been emitted, but value of the tables could be incorrect. I've updated the implementation to accumulate the changed tables.

I've updated the defaults for `throttleMs` and explained the behaviour in the comments.

simolus3 (approved):

PR #181, opened 2025-04-28
Add new schema options

simolus3:
This exposes the new table options introduced in version 0.3.13 of the core extension, namely:

- `include_old` (either a boolean or an array of column names): Adds an `old` entry to the `ps_crud.data` JSON object that includes old values for updates.
 - `include_old_only_when_changed`: Skip unchanged values from updates.
- `include_metadata`: Adds a `_metadata` column that can be written to for updates. It's reported through `CrudEntry.metadata` afterwards.
- `ignore_empty_update`: Don't create `ps_crud` entries for `UPDATE` statements that didn't change any values. 

Most of this feature is tested in the core extension, this adds a few simple tests ensuring we forward options correctly.


core/src/commonMain/kotlin/com/powersync/db/schema/Table.kt
```diff
@@ -33,6 +37,22 @@ public data class Table constructor(
      * Override the name for the view
      */
     private val viewNameOverride: String? = null,
+    /**
+     *  Whether to add a hidden `_metadata` column that will be enabled for updates to attach custom
+     *  information about writes that will be reported through [CrudEntry.metadata].
+     */
+    val includeMetadata: Boolean = false,
```
stevensJourney, line 44: This is just me blurbing out my initial thoughts on the naming of these variables.

The `include` here takes my brain a few cycles to understand what it means. The comment does explain it pretty well though. Maybe we could call this `trackMetadata`? Perhaps the same for `includeOld` -> `trackPreviousValues`?

core/src/commonMain/kotlin/com/powersync/db/schema/Table.kt
```diff
@@ -33,6 +37,22 @@ public data class Table constructor(
...
+     * When set to a non-null value, track old values of columns for [CrudEntry.oldData].
+     *
+     * See [IncludeOldOptions] for details.
+     */
+    val includeOld: IncludeOldOptions? = null,
+    /**
+     * Whether an `UPDATE` statement that doesn't change any values should be ignored when creating
+     * CRUD entries.
+     */
+    val ignoreEmptyUpdate: Boolean = false,
```
stevensJourney, line 55: Maybe call this `ignoreEmptyUpdates`?

simolus3:
cc @rkistner let me know what you think about the naming suggestions. I agree with them, but I'm not sure if we want a divergence between the JSON key internally and the public name? If that's ok I can apply the suggestion here and on the other PRs for this feature.

stevensJourney (commented):
This feature is pretty cool. I've left some minor naming comments, but I'm quite happy with the implementation.

Radiokot:
I'd love to have `ignoreEmptyUpdates`, thank you.

rkistner:
I like Steven's suggestions here.

It's fine if it diverges from the json, as long as it's consistent across the SDKs.

stevensJourney (approved):
Looks great to me 🚀

PR #182, opened 2025-04-29
CrudBatch hasMore + onChange improvement

stevensJourney:
# Overview

## CrudBatch

This fixes a bug where the `hasMore` value of `CrudBatch` was always false.

## onChange

This adds a `triggerImmediately` parameter to the `onChange` API similar to the [Dart API](https://github.com/powersync-ja/sqlite_async.dart/blob/6d85217e30c61c3e51372b2304aec6ae1f1660b6/packages/sqlite_async/lib/src/sqlite_queries.dart#L86). Adding this makes it easy to avoid race conditions in consumers such as the Swift SDK.

### DIFF
--- a/core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
+++ b/core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
@@ -187,14 +187,22 @@ class DatabaseTest {
     fun testTableChangesUpdates() =
         databaseTest {
             turbineScope {
-                val query = database.onChange(tables = setOf("users")).testIn(this)
+                val query =
+                    database
+                        .onChange(
+                            tables = setOf("users"),
+                        ).testIn(this)
 
                 database.execute(
                     "INSERT INTO users (id, name, email) VALUES (uuid(), ?, ?)",
                     listOf("Test", "test@example.org"),
                 )
 
-                val changeSet = query.awaitItem()
+                var changeSet = query.awaitItem()
+                // The initial result
+                changeSet.count() shouldBe 0
+
+                changeSet = query.awaitItem()
                 changeSet.count() shouldBe 1
                 changeSet.contains("users") shouldBe true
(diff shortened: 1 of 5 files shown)


core/src/commonIntegrationTest/kotlin/com/powersync/DatabaseTest.kt
```diff
@@ -418,4 +426,36 @@ class DatabaseTest {
...
+            }
+
+            // Purposely limit to less than the number of available ops
+            var batch = database.getCrudBatch(2) ?: error("Batch should not be null")
+            batch.hasMore shouldBe true
+            batch.crud shouldHaveSize 2
+            batch.complete(null)
+
+            batch = database.getCrudBatch(1000) ?: error("Batch should not be null")
+            batch.crud shouldHaveSize 1
```
simolus3, line 456: Should we assert that `hasMore` is not true here?

simolus3 (dismissed):
The CRUD changes look good to me.

> Adding this makes it easy to avoid race conditions in consumers such as the Swift SDK.

What race conditions are we aiming to avoid with this? It's not immediately clear to me, so maybe you have an example where it helps that we have an initial emit right away?

stevensJourney:
> The CRUD changes look good to me.
> 
> > Adding this makes it easy to avoid race conditions in consumers such as the Swift SDK.
> 
> What race conditions are we aiming to avoid with this? It's not immediately clear to me, so maybe you have an example where it helps that we have an initial emit right away?

In Swift we'd like to use the `onChange` method to implement the `watch` method. Doing this gives the Swift SDK the ability to control and catch any exceptions which might occur when performing the corresponding `getAll` query. 

It can become tricky to control the emission of the first watched query result and starting a Task which listens to events from an `onChange` stream and then executes a `getAll` updating the resultant stream. It's not impossible to manage this, but having the onChange method emit an initial event makes the implementation much simpler.

PR #184, opened 2025-04-30
[Internal] Swift Exception Helper

stevensJourney:
# Overview

We explicitly annotate certain methods in the Kotlin SDK to indicate that they throw `PowerSyncException`s. This allows the Swift SDK to catch these exceptions when a Kotlin method is invoked from Swift.

Catching exceptions in Swift is easy, but it can be tricky to throw Exceptions which should be handled by the Kotlin SDK from the Swift SDK. One cannot throw the SKIEE translated `PowerSyncException`s from Swift code.

This PR introduces a small helper which throws a `PowerSyncException` from Kotlin code. In some situations (unfortunately not all) this makes reporting exceptions from Swift to the Kotlin SDK simple.

See the usage in the Swift SDK:
[Here](https://github.com/powersync-ja/powersync-swift/blob/223c004be120b71382a89c211a162ae647538569/Sources/PowerSync/Kotlin/wrapQueryCursor.swift#L76)
[Here](https://github.com/powersync-ja/powersync-swift/blob/223c004be120b71382a89c211a162ae647538569/Sources/PowerSync/kotlin/PowerSyncBackendConnectorAdapter.swift#L35)


core/src/commonMain/kotlin/com/powersync/PowerSyncException.kt
```diff
@@ -4,3 +4,16 @@ public class PowerSyncException(
...
+ * Helper class designed to bridge SKIEE methods and allow them to throw
+ * `PowerSyncException`. This is necessary because these exceptions cannot
+ * be thrown directly in Swift.
+ *
+ * The class provides a mechanism to handle exceptions in a way that is
+ * compatible with the Swift environment, ensuring proper error propagation
+ * and handling.
+ */
+@Throws(PowerSyncException::class)
+public fun throwPowerSyncException(exception: PowerSyncException): Unit = throw exception
```
simolus3, line 19: Since this will only be used for Swift and nowhere else, can we move this into the `PowerSyncKotlin` project? My understanding is that sources from there would also be considered for SKIEE (but maybe do a local build to validate, if it doesn't work we can keep it here).

simolus3 (commented):
LGTM, just want to check if we can move this to `:PowerSyncKotlin` to hide this from non-Swift clients.

core/src/commonMain/kotlin/com/powersync/PowerSyncException.kt
stevensJourney: That is a very good suggestion! It looks like we can move this to `PowerSyncKotlin`. I did a clean build and was able to use this method in the Swift SDK 🙌

simolus3 (approved):

PR #194, opened 2025-05-21
Refactor SQLite compilation tasks

simolus3:
On all native platforms we currently support, we compile and link a copy of SQLite statically (through the `:static-sqlite-driver` project).

The way the compilation is currently set up is a suboptimal, as it doesn't allow Gradle to cache outputs effectively:

1. We're using `dependsOn` to specify task dependencies, which is [not recommended](https://docs.gradle.org/current/userguide/best_practices_tasks.html#avoid_depends_on). We should specify file dependencies instead, as that allows Gradle to understand the dependencies better and allows for fine-grained build invalidation.
2. A lot of the compilation tasks we written using `Exec`, which does not support caching. To make the build easier to understand while also improving cache invalidation, I've extracted the tasks into independent files.

Finally, I've enabled the [build cache](https://docs.gradle.org/current/userguide/build_cache.html), which should allow us to cache outputs (including the compiled sqlite libraries) across CI runs, avoiding a recompilation on every run.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
+++ b/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
@@ -436,8 +436,8 @@ internal class SyncStream(
         state: SyncStreamState,
     ): SyncStreamState {
         val batch = SyncDataBatch(listOf(data))
-        status.update { copy(downloading = true, downloadProgress = downloadProgress?.incrementDownloaded(batch)) }
         bucketStorage.saveSyncData(batch)
+        status.update { copy(downloading = true, downloadProgress = downloadProgress?.incrementDownloaded(batch)) }
         return state
     }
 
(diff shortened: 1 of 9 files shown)


core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
```diff
@@ -436,8 +436,8 @@ internal class SyncStream(
         state: SyncStreamState,
     ): SyncStreamState {
         val batch = SyncDataBatch(listOf(data))
-        status.update { copy(downloading = true, downloadProgress = downloadProgress?.incrementDownloaded(batch)) }
         bucketStorage.saveSyncData(batch)
+        status.update { copy(downloading = true, downloadProgress = downloadProgress?.incrementDownloaded(batch)) }
```
simolus3, line 440: I think moving this down here may improve test reliability.

In particular with the two "sync progress after interrupted sync" tests, what could happen is that we update the state here, the test asserts that the state was updated and immediately closes the database (before we're able to save the batch). Then when we connect again, the test asserts that some data has already been downloaded but it's not the case due to us closing the database too early.

So let's hope that this improves the flaky tests we've seen 🤞

stevensJourney (approved):

PR #195, opened 2025-05-21
WIP: Try enabling configuration cache

simolus3:
Using the [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) can speed up builds by caching results of the configuration phase, which means that Gradle doesn't have to evaluate buildscripts multiple times.

An even bigger impact is that enabling this allows tasks in the same project to run in parallel, which is particularly helpful when working on the SDK: The steps to compile SQLite for `:static-sqlite-driver` are sped up a lot by running them in parallel (by default, we'd compile SQLite one ABI at a time).

I've heard that the configuration cache doesn't work too well with some publishing plugins, so I've disabled it for the publishing CI. Apart from that, this should hopefully speed up our builds. We could cache the results of the configuration cche between CI runs too by [setting a key to encrypt them](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#saving-configuration-cache-data) with (it's not cached by default because the data may be sensitive). But I think the benefits of multi-threading alone could justify this and we don't necessarily need to worry about that.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -35,12 +35,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: ~/.konan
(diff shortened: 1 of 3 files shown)


stevensJourney (approved):
Looks good to me :)

PR #203, opened 2025-06-19
Include version number in user agent, prepare release

simolus3:
This adds a simple Gradle task generating a source set with a single file containing the version of the SDK. This is then imported when creating the user agent to always include the version.

Apart from that, this raises the version to 1.2.0 and prepares a release.


stevensJourney (dismissed):
Nice! LGTM!

stevensJourney (approved):

PR #205, opened 2025-06-22
Update error handling to capture all 4xx responses

### LINKED ISSUE #204: supabase fatal errors
the supabaseconnector is only checking for 400 error codes [response.status.value **==** 400](https://github.com/powersync-ja/powersync-kotlin/blob/main/connectors/supabase/src/commonMain/kotlin/com/powersync/connector/supabase/SupabaseConnector.kt#L96).
shouldn't it be **>=** 400, for example to catch fatal error  [INSUFFICIENT PRIVILEGE 42501](https://github.com/powersync-ja/powersync-kotlin/blob/main/connectors/supabase/src/commonMain/kotlin/com/powersync/connector/supabase/SupabaseConnector.kt#L50)? which is sent back in status code 403.

fethij:
interceptor was only capturing error codes when the HTTP status was exactly 400, but RLS violations from Supabase were coming with a different HTTP status code. For example fatal error [INSUFFICIENT PRIVILEGE 42501](https://github.com/powersync-ja/powersync-kotlin/blob/main/connectors/supabase/src/commonMain/kotlin/com/powersync/connector/supabase/SupabaseConnector.kt#L50) was 403.

closes #204

### DIFF
--- a/connectors/supabase/src/commonMain/kotlin/com/powersync/connector/supabase/SupabaseConnector.kt
+++ b/connectors/supabase/src/commonMain/kotlin/com/powersync/connector/supabase/SupabaseConnector.kt
@@ -93,7 +93,7 @@ public class SupabaseConnector(
         supabaseClient.httpClient.httpClient.plugin(HttpSend).intercept { request ->
             val resp = execute(request)
             val response = resp.response
-            if (response.status.value == 400) {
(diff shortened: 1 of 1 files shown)


stevensJourney (approved):
Thanks! This looks good to me. I'll add a changelog entry in a follow-up PR.

Issue #219, opened 2025-07-12
PowerSyncBackendConnector.fetchCredentials() error crashes an Android app


Radiokot:
Hey guys,

So the docs for this method say:
> Return null if the user is not signed in. Throw an error if credentials cannot be fetched due to a network error or other temporary error.

In a situation when the device has been offline long enough for a Supabase token to expire and stays offline when you visit the app, the `fetchCredentials()` method throws `PowerSyncException: Supabase client is not authenticated` because the expired token can't be renewed. I think the app should continue to work in this case, but sadly it crashes (FATAL EXCEPTION).

My PowerSync version is `1.2.1`, with the stable client.

Thanks.

PR #220, opened 2025-07-16
Refactor fetching credentials

simolus3:
In the `PowerSyncBackendConnector` implementation, errors thrown in `fetchCredentials()` are currently swallowed silently, because:

1. We use a supervisor scope to launch jobs fetching credentials.
2. We then call `join()` on a `Job` instead of using a `Deferrable` and `await()`. `join()` just waits for the job to finish and, for supervisor scopes, doesn't rethrow exceptions.
3. So, `getCredentialsCached()` would silently return `null` on errors, without this information being surfaced to the user.

On Android, it also looks like errors in root coroutines that aren't handled will unconditionally kill the app, regardless of the supervisor scope present (I can reproduce https://github.com/powersync-ja/powersync-kotlin/issues/219 by simply throwing in a connector's `fetchCredentials` on Android).

This PR refactors the implementation to avoid a global coroutine scope. This scope, with a `Job` to fetch credentials essentially had two purposes: First, to avoid fetching credentials concurrently (we want to re-use results instead), and second to implement prefetching credentials asynchronously.
For the first requirement, using a proper mutex is both easier and safer. For the second requirement, it's much better to let callers bring their own `CoroutineScope` that will properly forward errors. We are indeed already using `scope.launch` for the Rust client (the only place prefetching credentials) - so the internal async management can just be removed. This ensures errors in `fetchCredentials` bubble up all the way to the `try/catch` in the sync client which will retry after a delay.

Introducing a final method makes `PowerSyncBackendConnector` impossible to mock, so I've migrated tests to use the `TestConnector` based on callbacks instead.

Closes https://github.com/powersync-ja/powersync-kotlin/issues/219.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Support tables created outside of PowerSync with the `RawTable` API.
   For more information, see [the documentation](https://docs.powersync.com/usage/use-case-examples/raw-tables).
 * Fix `runWrapped` catching cancellation exceptions.
+* Fix errors in `PowerSyncBackendConnector.fetchCredentials()` crashing Android apps.
 
 ## 1.2.2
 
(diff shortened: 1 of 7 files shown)


Radiokot:
Thank you!

PR #222, opened 2025-07-16
feature: getNextCrudTransactionBatch

whygee-dev:


core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
```diff
@@ -134,6 +134,25 @@ public interface PowerSyncDatabase : Queries {
...
+     * requesting the next batch.
+     *
+     * Unlike [getCrudBatch], this groups data by transaction, allowing developers to
+     * upload multiple complete transactions in a single batch operation.
+     *
+     * @param transactionLimit The maximum number of transactions to include in the batch.
+     * Default is 5.
+     */
+    @Throws(PowerSyncException::class, CancellationException::class)
+    public suspend fun getNextCrudTransactionBatch(transactionLimit: Int = 10): CrudBatch?
```
simolus3, line 154: Another thing we would likely want here is to have an optional limit on the amount of crud items - here it can essentially be unbounded if you have multiple huge transactions.

I'm not sure yet if we'll need both limits or if a limit on items is enough. Basically the behavior would be to return the maximum amount of transactions so that the total items in those transactions doesn't exceed the limit, but also return at least one transaction.

core/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -309,6 +309,62 @@ internal class PowerSyncDatabaseImpl(
         }
     }
 
+    override suspend fun getNextCrudTransactionBatch(transactionLimit: Int): CrudBatch? {
+        waitReady()
+        return internalDb.readTransaction { transaction ->
+            // Since tx_id can be null, we can't use a WHERE tx_id < ? with transactionLimit + first crud entry tx_id
+            // So we get all operations and group them by transaction or fall back to an individual transaction if tx_id is null
+            val allOperations =
```
simolus3, line 317: We would likely need an optimized query here that does the counting in SQL - fetching all items first could be too inefficient.

I know this isn't easy to get right in SQL. Some windowing magic might help, if you need suggestions we're also happy to help.

simolus3 (commented):
Thank you for your contribution!

We still think this is a feature we want to have, but after some discussion we also want to get the parameters for limits right and want to ensure this method performs well even with a large amount of local crud items. Another thing we're looking at is possibly adding an index on `ps_crud.tx_id` to make queries faster (but we might be able to rely on the fact that the rowid order matches the transaction order - something to take a look at depending on what query we use for this in the end).

simolus3:
Sorry for the delay on this. We've discussed this internally, and the API we want is to expose transactions as a `Flow`. This means that developers get transactions one-by-one, and can decide on their own when they have enough (instead of an approach like this providing a static limit).

Still, thank you for getting the ball rolling on this and your contribution! I've opened https://github.com/powersync-ja/powersync-kotlin/pull/245 with the new API based on the discussion we had here.

PR #228, opened 2025-07-21
Upgrade kotlinx-datetime to 0.7.1

ecstatic-morse:
This involves migrating to the [builtin versions of a few types](https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#deprecation-of-instant):

-  `kotlinx.datetime.Instant` -> `kotlin.time.Instant`
-  `kotlinx.datetime.Clock` -> `kotlin.time.Clock`

These are still experimental, so add the corresponding `optIn` flag as well.


simolus3:
Actually it looks like we'll have a hard time with this:

1. Versions of the supabase-kt project older than 3.2.1 don't support the datetime migration yet (there are runtime failures about missing classes - `java.lang.ClassNotFoundException: kotlinx.datetime.serializers.InstantIso8601Serializer`).
2. For me, updating supabase-kt to 3.2.1 resulted in compiler errors, probably because we need to upgrade our Kotlin version too.
3. However, we're currently blocked on updating Kotlin due to https://github.com/touchlab/SKIE/issues/162

Is this blocking something for you?

ecstatic-morse:
Not a blocker. This was just the last of my dependencies using version 0.6.x of `kotlinx-datetime`.

I mistakenly thought this only required `kotlin` >2.1.20 and `kotlinx-serialization` >1.9.0, but  it looks like there's a bit more [going on](https://github.com/Kotlin/kotlinx.serialization/issues/3026#issuecomment-3018712811).

Feel free to put this on the pause until things settle down a bit.

Issue #208, opened 2025-06-26
New sync client doesn't recover from a connection failure on Android


Radiokot:
Hey guys, so I briefly tried `newClientImplementation` and found out that it doesn't recover from a connection failure, meaning it doesn't retry uploading the transaction if the first attempt failed due to missing data connection. I also tried switching `newClientImplementation` and `method` individually, and the issue occurs only with `newClientImplementation = true` so apparently nothing to do with `ConnectionMethod.WebSocket`.

To reproduce:
- Write to the database while offline
- Wait for the sync attempt to fail
- Enable the data connection
- Wait for the changes to appear in Supabase

With the stable client, you have the following:
```
... do the write ...
Error in streamingSync: Unable to resolve host
... enable data connection ...
Could not apply checkpoint. Waiting for next sync complete line
... Supabase connector is called, transaction uploaded ...
validated checkpoint Checkpoint(...)
```

With the new client, you have this:
```
... do the write ...
Error in streamingSync: Unable to resolve host
... enable data connection ...
Could not apply checkpoint. Waiting for next sync complete line
[SQLite EXEC] BEGIN TRANSACTION
powersync_control(line_text, {"token_expires_in":602356})
[SQLite EXEC] SELECT powersync_control(?, ?) AS r
control result: []
... nothing else happens ...
```

Force-closing and re-opening the app doesn't help. The only way to get the transaction uploaded is to make another one, on this exact device with the data connection enabled – then the sync kicks back and the failed transaction gets uploaded.

Thanks.

simolus3:
Version 1.2.2 should fix this issue for websockets as well, thanks again for the follow-up!

Radiokot:
Are you sure?

Radiokot:
Got it, I'll retest myself once the fix is released and let you guys know if the sync recovers or not

Radiokot:
If you do this in Android Studio Koala, where embedded Kotlin plugin doesn't support 2.2.0, everything becomes red, although the prpject builds successfully with Gradle.

PR #232, opened 2025-07-30
Remove websocket transport, use binary HTTP stream when available

simolus3:
There have been a few issues with websocket streams that are tricky for us to fix. In particular, one issue we've seen is that the sync service [sends lines exceeding the default message size for Darwin clients](https://discord.com/channels/1138230179878154300/1400010537504346163/1400010537504346163), causing connection resets. Worse, due to [this KTOR bug](https://youtrack.jetbrains.com/issue/KTOR-6963/Darwin-The-maxFrameSize-option-has-no-effect), we can't even customize the frame size, making the WebSocket transport mode fundamentally unstable on Apple platforms.

So, this removes the experimental option for websockets from the SDK. Instead, it adopts the option for [binary HTTP streams](https://github.com/powersync-ja/powersync-service/pull/310) in the sync service. When both client and sync service are compatible with this, they'll send BSON data over the HTTP stream. That way, we have most of the benefits of the websocket protocol ported to HTTP.

This also configures the socket timeout option to twice the usual frequency used by the sync service, allowing us to detect broken HTTP connections.

Closes https://github.com/powersync-ja/powersync-kotlin/issues/208.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
+++ b/core/src/commonMain/kotlin/com/powersync/sync/SyncStream.kt
@@ -62,8 +62,6 @@ import kotlinx.io.readIntLe
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.encodeToJsonElement
-import kotlin.time.Duration
-import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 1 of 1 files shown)


stevensJourney (approved):
I'm happy with the changes here, especially given the current issues with WebSockets on Apple platforms.

PR #239, opened 2025-08-05
Typed crud values

simolus3:
Most of our SDKs (I've checked Dart and JS) represent `CrudEntry.opData` and `CrudEntry.previousValues` as a `Map<String, Any?>`. In Kotlin however, we're using a `Map<String, String?>` ([reported on Discord](https://discord.com/channels/1138230179878154300/1400787677875408907/1400787677875408907)). This is somewhat unfortunate, as:

1. It means that the Supabase connector will upload everything as a string, regardless of the actual type. While this isn't a huge problem since Postgrest will cast, it's also not great.
2. It's a blocker if we want to represent binary data in the future.
3. Since SQLite is dynamically typed, it would be nice to properly preserve type information so that users can act on the type of values at runtime.

This PR adds information to construct the type of values in `opData` and `previousValues` with backwards-compatibility in mind:

1. As a technically-breaking change, the `CrudEntry` constructor is made `internal`. Users probably weren't invoking it directly though, that's an SDK responsibility.
2. To keep the public `Map<String, String?>` type, this introduces the `SqliteRow` interface extending that type, while also providing a getter to get access to typed values.
3. The interface also has a method to return a `JsonObject`, which we can use in the Supabase connector to upload values unchanged (giving users a serializable object without making `SqliteRow` serializable, which would blow up the framework size).

This way, existing reads on the fields still work because we have a `Map<String, String?>` type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 * Added the ability to log PowerSync service HTTP request information via specifying a
   `SyncClientConfiguration` in the `SyncOptions.clientConfiguration` parameter used in
   `PowerSyncDatabase.connect()` calls.
-* `CrudEntry`: Add `data` and `typedPreviousValues` fields as typed variants of
-  `opData` and `previousValues`, respectively.
+* `CrudEntry`: Introduce `SqliteRow` interface for `opData` and `previousValues`, providing typed
+  access to the underlying values.
 
 ## 1.3.1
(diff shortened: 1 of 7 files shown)


stevensJourney (approved):
Happy with the changes here. This looks like a good solution which greatly improves the DX while maintaining backwards compatibility.

PR #242, opened 2025-08-07
Fix some logging issues

Radiokot:
Hey guys,

This pull request addresses some logging issues I spotted long ago, but today I wanted to have certain PowerSync logs in my common app log, and this was time to fix them.

1. Fixed modifying severity of the global Kermit logger – it is up to the app owner to decide on this parameter. The library no longer uses the global instance by default;
2. Added `PowerSync` tag to the logs – as now the library has its own default logger instance, it deserves a tag;
3. Removed `generateLogger()` global public method – it really seemed like a method meant to be used internally as it is quite specific, but even internally it is no longer needed.

I didn't run any tests because I couldn't wait for the project to build. On my Intel Core Ultra 7 14x4.8 GHz with 32 GB of RAM the **build didn't finish in 45 minutes**, it's already late here and I should go to bed soon.


simolus3 (dismissed):
I agree with these changes, and looking at an example app it's nice that we now have a proper tag for PowerSync-related logs 👍 

> I didn't run any tests because I couldn't wait for the project to build. On my Intel Core Ultra 7 14x4.8 GHz with 32 GB of RAM the build didn't finish in 45 minutes, it's already late here and I should go to bed soon.

That's probably an issue with our build setup and not with your machine. Do you remember a specific task taking too long (or was it just stuck downloading dependencies)? We build or SDK on GH actions which is a lot slower but still seems to work. So I assume there may be something locking up here.

Radiokot:
> Do you remember a specific task taking too long (or was it just stuck downloading dependencies)?

It was just the initial Gradle sync, if I remember correctly it is called "Build model". It was not stuck, it just kept downloading and building, and so on and so on.

stevensJourney (approved):

PR #249, opened 2025-08-22
Add Room integration

simolus3:
This adds a Room integration package for PowerSync.

This works by making the PowerSync SDK run queries against the connection pool managed for Room databases, which ensures that we have no conflict due to multiple write connections. 

This uses the new update hooks implemented in the core extension to query for updates after each transaction from the PowerSync SDK. These updates are automatically forwarded to Room (so a sync from PowerSync automatically updates `Flow`s in Room).
Because we can't install a general update listener on Room databases, the other direction is more complicated. The best approach I could come up with here is to create a Room flow listening on all tables explicitly, and then call `powersync_update_hook('get')` to forward these writes to PowerSync.

One remaining issue is that the driver APIs exposed by Room are async-only. Using `runBlocking` to turn them works, but is a bit hacky and not particularly efficient. I will open an API for asynchronous transactions as a medium-term workaround.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           - os: windows-latest
             targets: jvmTest
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 30
 
     steps:
     - uses: actions/checkout@v4


integrations/room/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+powersync.connect(...)
+```
+
+Changes from PowerSync (regardless of whether they've been made with `powersync.execute` or from a
+sync operation) will automatically trigger updates in Room.
+
+To also transfer local writes to PowerSync, you need to
+
+1. Create triggers on your Room tables to insert into `ps_crud` (see the PowerSync documentation on
```
jgillich, line 51: Would be nice to link the docs. [This](https://docs.powersync.com/usage/use-case-examples/raw-tables#capture-local-writes-with-triggers)?

simolus3: I agree, thanks! We'll probably also have a documentation page specifically for Room with more examples.

core/src/watchosMain/kotlin/com/powersync/DatabaseDriverFactory.watchos.kt
```diff
@@ -31,3 +31,9 @@ private val didLoadExtension by lazy {
 
     true
 }
+
+@ExperimentalPowerSyncAPI
+public actual fun resolvePowerSyncLoadableExtensionPath(): String? {
+    didLoadExtension
```
stevensJourney, line 37: It looks like this method can also throw if the extension has not been loaded correctly. It seems like the JVM version also throws in this case. Should we annotate the `resolvePowerSyncLoadableExtensionPath` method with `@Throws`?

integrations/room/README.md
```diff
@@ -0,0 +1,62 @@
...
+    val driver = BundledSQLiteDriver().also {
+        it.loadPowerSyncExtension() // Extension method by this module
+    }
+    
+    Room.databaseBuilder(...).setDriver(driver).build()
+    ```
+3. Configure raw tables for your Room databases.
+
+After these steps, you can open your Room database like you normally would. Then, you can use the
+following method to obtain  a `PowerSyncDatabase` instance that is backed by Room:
```
stevensJourney, line 31: Nit: I think this is better (I am terrible with grammar though) 
```suggestion
following method to obtain  a `PowerSyncDatabase` instance which is backed by Room:
```

integrations/room/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```Kotlin
+val pool = RoomConnectionPool(yourRoomDatabase)
+val powersync = PowerSyncDatabase.opened(
+    pool = pool,
+    scope = this,
+    schema = Schema(...), // With Room, you need to use raw tables
```
stevensJourney, line 38: Nit: Maybe add a small mention why this is the case and link to the general Raw table [docs](https://docs.powersync.com/usage/use-case-examples/raw-tables) early (I see the Raw table writes are linked further below)

integrations/room/README.md
```diff
@@ -0,0 +1,62 @@
...
+To also transfer local writes to PowerSync, you need to
+
+1. Create triggers on your Room tables to insert into `ps_crud` (see the
+   [PowerSync documentation on raw tables](https://docs.powersync.com/usage/use-case-examples/raw-tables#capture-local-writes-with-triggers)
+   for details).
+2. Listen for Room changes and invoke a helper method to transfer them to PowerSync:
+    ```Kotlin
+    yourRoomDatabase.getCoroutineScope().launch {
+        // list all your tables here
+        yourRoomDatabase.invalidationTracker.createFlow("users", "groups", /*...*/).collect {
```
stevensJourney, line 58: Could we do this automatically, internally using the provided Schema's tables?

stevensJourney (approved):
Left some questions, but overall this looks very cool.

integrations/room/README.md
simolus3: The pool implementation does not have access to the schema, but I've added a method to invoking this snippet for a given `Schema` instance (and also added a constructor parameter to automatically call that method).

stevensJourney (dismissed):
🚀

PR #252, opened 2025-08-25
Fix null values in CRUD entries being reported as strings

simolus3:
When a column in a CRUD entry was set to `NULL`, the Kotlin SDK currently reports that as `"null"` instead of the actual `null` value. This PR fixes that problem.


stevensJourney (dismissed):
Thanks for the quick fix!

stevensJourney (approved):

PR #258, opened 2025-09-10
Support asynchronous transactions

simolus3:
Since Room uses an asynchronous-only connection pool, integrating it into the PowerSync SDK would be more efficient if we didn't have to call `runBlocking` around every statement.
Making query methods asynchronous also simplifies some interfaces, since it means that the top-level `Queries` and inner transactions can share asynchronous methods.

So, this PR:

1. Introduces the `ScopedReadQueries` and `ScopedWriteQueries` interfaces. The `get`, `getAll` and `getOptional` methods have been moved to `ScopedReadQueries`; `execute` is now part of `ScopedWriteQueries`.
2. The `Queries` interface now extends from `ScopedWriteQueries`.
3. Instead of implementing virtually every method twice (once in `PowerSyncDatabaseImpl` to await for initialization and then in `InternalDatabaseImpl`), most methods are now implemented directly in `Queries` ontop of the `useConnection` interface.
  - This doesn't break waiting for initialization, as `useConnection` does that for all methods.

I have marked the synchronous callbacks as `@Deprecated`. I'm not sure if we want to be that aggressive or if we want to keep supporting both methods for longer. I'm slightly leaning toward eventually removing them given the nicer unified interface and reduced complexity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Remove internal SQLDelight and SQLiter dependencies.
 * Add `rawConnection` getter to `ConnectionContext`, which is a `SQLiteConnection` instance from
   `androidx.sqlite` that can be used to step through statements in a custom way.
+* Support asynchronous transactions and locks.
 
 ## 1.5.1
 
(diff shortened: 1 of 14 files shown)


simolus3:
Closing in favor of https://github.com/powersync-ja/powersync-kotlin/pull/354.

PR #260, opened 2025-09-16
Add helpers for Swift 6.2 compatibility

simolus3:
In Swift 6.2, released with XCode 26, the Swift implementation of the backend connector fails to compile. The reason is that it receives a database instance through the callback, which we can't properly mark as `Sendable`.
This PR adds a Kotlin workaround for that by exposing a connector interface that doesn't need that parameter.

Additionally, this fixes `getCrudTransactions()` to not be `suspend` (it returns a `Flow` so it doesn't have to be) and adds a helper exposing that flow as a sequence of results (because flows can't throw with SKIE). This will allow us to add `getCrudTransactions()` to the Swift SDK.

This will be adopted in Swift with https://github.com/powersync-ja/powersync-swift/pull/79


stevensJourney (approved):
Happy with these changes

stevensJourney (approved):

PR #261, opened 2025-09-16
Make CI more reliable

simolus3:
This refactors the `watch()`, `onChange()` and `AttachmentTest` implementations to improve the reliability of the CI.

First, both `watch()` and `onChange()` were implemented as a `channelFlow` this is a somewhat problematic builder, since (quoting [the docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/channel-flow.html)):

>A channel with the [default](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/-factory/-b-u-f-f-e-r-e-d.html) buffer size is used.

This means that when we call `send()`, it doesn't necessarily wait for the downstream consumer to process that event, leading to the query potentially running more often than necessary (the behavior we want is to internally buffer upstream changes while a downstream consumer is busy, and then once it handled the previous event we'd re-run the query if there were any changes in the meantime).
I've refactored the implementation to use a regular flow, ensuring that we won't start running a second query until results of the first one are handled. I've also simplified the implementation to reuse more logic between `watch` and `onChange`.

Additionally, this updates some attachment tests:

- The `testSkipFailedDownload` test asserts a visible transition from `QUEUED_DOWNLOAD` to `ARCHIVED`. Depending on how often the `SELECT * FROM attachments` flow emits items though, it's possible that the `QUEUED_DOWNLOAD` state is not observable. This updates the test to reflect that.
- Because the actual attachment sync and the observed `SELECT * FROM atttachments` flow are decoupled and asynchronous, it's possible that a write to the todos table schedules a query that, at the time it runs, sees a different write. This has been causing a few failures as well, a `distinctUntilChanged()` deflakes those tests.
- I also noticed some failures related to `waitFor` timing out, so this adds more logs around that and replaces the `waitFor` calls in attachment tests.

### DIFF
--- a/core/src/commonIntegrationTest/kotlin/com/powersync/AttachmentsTest.kt
+++ b/core/src/commonIntegrationTest/kotlin/com/powersync/AttachmentsTest.kt
@@ -58,7 +58,7 @@ class AttachmentsTest {
 
     private fun ActiveDatabaseTest.watchAttachmentsTable(): Flow<List<Attachment>> =
         database
-            .watch("SELECT * FROM attachments -- test") {
+            .watch("SELECT * FROM attachments") {
                 Attachment.fromCursor(it)
             }
             // Because tests run on slow machines, it's possible for a schedule like the following


stevensJourney (approved):
This must have taken sine effort to find and debug. Thanks for this! Looks good to me.

PR #267, opened 2025-09-20
Improve room integration docs

jgillich:
* Added raw table example
* Set required option `newClientImplementation=true`


jgillich:
Ah [the docs page](https://docs.powersync.com/client-sdk-references/kotlin-multiplatform/libraries/room) page is much better than the current readme. Perhaps it's best to remove all instructions from the readme

PR #270, opened 2025-09-23
Sync streams demo

simolus3:
This adds:

1. Stream compose hooks (if that's what they're called) to the `:compose` project.
2. A new desktop-only (for simplicity) todolist demo talking to the node demo backend with sync streams and SQLDelight.

The new demo is read-only apart from the fact that items can be marked as complete / incomplete. There's no functionality to create new lists or items, although that could be added later.


simolus3:
Closing in favor of https://github.com/powersync-ja/powersync-kotlin/pull/329.

PR #273, opened 2025-09-29
Add in-memory database for tests

simolus3:
This adds the `PowerSyncDatabase.inMemory` factory to create PowerSync databases backed by a single in-memory connection.

This utility simplifies the test setup for users:

1. It doesn't need an open factory, so users can write tests in `commonMain` instead of having to setup `expect` / `actual`s in tests.
2. One doesn't have to create a temporary file path for each test, databases are isolated on their own.
3. No cleanup logic is required.

We should continue testing the core logic with "real" file-backed databases to ensure the pool works, but having a single in-memory connection is useful to test other packages that can expect `:core` to behave correctly. This adopts that for the SQLDelight integration, I'm also writing tests for the Supabase connector where these changes would simplify the setup.

### DIFF
--- a/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
+++ b/core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
@@ -10,10 +10,12 @@ import com.powersync.db.Queries
 import com.powersync.db.crud.CrudBatch
 import com.powersync.db.crud.CrudTransaction
 import com.powersync.db.driver.SQLiteConnectionPool
+import com.powersync.db.driver.SingleConnectionPool
 import com.powersync.db.schema.Schema
 import com.powersync.sync.SyncOptions
 import com.powersync.sync.SyncStatus
 import com.powersync.utils.JsonParam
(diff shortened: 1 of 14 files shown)


core/src/commonMain/kotlin/com/powersync/PowerSyncDatabase.kt
```diff
@@ -233,6 +235,29 @@ public interface PowerSyncDatabase : Queries {
...
+        @OptIn(ExperimentalPowerSyncAPI::class)
+        public fun inMemory(
+            schema: Schema,
+            scope: CoroutineScope,
+            logger: Logger? = null,
+        ): PowerSyncDatabase {
+            val logger = generateLogger(logger)
+            // Since this returns a fresh in-memory database every time, use a fresh group to avoid warnings about the
+            // same database being opened multiple times.
+            val collection = ActiveDatabaseGroup.GroupsCollection().referenceDatabase(logger, "test")
```
stevensJourney, line 250: If we're creating a fresh group, should the identifier not be unique here?

PR #274, opened 2025-09-29
Supabase error handling

simolus3:
This expands the supabase connector to make it `open` and allow a few methods to be customized:

- Uploading individual crud rows has been extracted into an `open` method.
- Handling errors is also an `open` method.

This is to address [this reported issue](https://discord.com/channels/1138230179878154300/1416775559542866023/1416775559542866023) where a user wants to report errors to Crashlytics. At the moment, because the connector is final and catches fatal Postgres exceptions by (apart from a logger, essentially silently) deleting local writes, that would be tricky to detect (it doesn't even show up as an upload error in the sync status).

To clean up some top-level directories, this also moves the Supabase connector into the `integrations/` folder. The maven artifact id is not affected by this because of a `gradle.properties` file with a `POM_ARTIFACT_ID` entry.


stevensJourney (dismissed):
This is a great improvement to the current implementation! I left some minor wording suggestions, but this LGTM.

stevensJourney (dismissed):

stevensJourney (approved):

PR #275, opened 2025-09-29
Move `PowerSyncKotlin/` to `internal/`

simolus3:
Having a directory named `PowerSyncKotlin/` in the root of the repository is somewhat confusing: It makes it sounds like that's where the PowerSync Kotlin SDK is, when in reality it's mostly unrelated to the Kotlin SDK.

Since we've recently introduced the `internal/` folder for another subproject, that sounds like a good place to put the Swift SDK helper to tidy up the workspace a bit.


stevensJourney (commented):
I think we'd still need to update the `deploy.yml` actions for this new structure?

simolus3:
Good catch, thanks! Done.

PR #277, opened 2025-10-17
Encryption part 2: Use common SDK for Swift build

simolus3:
This migrates the Swift helpers to use the `:common` package instead of `:core`, meaning that sqlite3 will no longer be linked into the created framework.

This adds helpers to `:common` making it easier to replicate the default connection factory from `:core`. This is now used in the Swift helper to add a function returning such as a connection factory. The connection factory can be customized with statements to run when the connection is initialized. This is useful to e.g. set a `PRAGMA key = `.

This will be used in the Swift SDK like shown in this PR: https://github.com/powersync-ja/powersync-swift/pull/82. To unblock Swift after these changes, see https://github.com/powersync-ja/powersync-swift/pull/83.


stevensJourney (approved):
This looks good to me :D

PR #288, opened 2025-11-10
Add encryption support through SQLite3MultipleCiphers

simolus3:
This adds the `com.powersync:sqlite3multipleciphers` package, providing encryption support.

This package has to replicate what `:core` is using as a SQLite driver on all targets:

- For Kotlin/Native, we replicate what `static-sqlite3-driver` does already (except that we link SQLite3MultipleCiphers instead of the normal SQLite library). Since these are ABI-compatible, we can use the native driver that is already part of the `:common` package.
- For Android and JVM, the situation is a bit trickier. Since there is no good pre-packaged build of SQLite3MultipleCiphers (and since SQLCipher has a limited API that can't be used to implement our driver API), we need our own build.
  - The package is essentially a copy of what `androidx.sqlite:bundled-sqlite-driver` does. I've copied the sources and kept copyright headers. The only change is that the package name has been updated to `com.powersync.encryption`. The sources contain Kotlin classes and the `sqlite_bindings.cpp` file providing a JNI wrapper around SQLite.
  - Cross-compiling JNI libraries is madness and took me several approaches. We run the compilation on macOS hosts only, and:
    - For macOS targets, use the clang installation from XCode.
    - For Windows targets, use [mingw-w64](https://www.mingw-w64.org/). Except that doesn't have prebuilt binaries that work on Apple silicon macs, so we're actually using [llvm-mingw](https://github.com/mstorsjo/llvm-mingw).
    - For Linux targets, use `clang` from a homebrew installation already installed on GitHub assets. Clang can cross-compile natively, but will need libc headers and compiler builtins for the target platform. I've added a shellscript that downloads these from Archlinux packages, the Gradle build task then points clang towards the relevant sysroot.

I really don't like this build process, but at least it runs reliably in CI. I've opened [a pull request](https://github.com/powersync-ja/powersync-kotlin/pull/300) to attach the output of this build to GitHub releases doing forward. After the next release, I'll configure the build to download assets from there by default.

There's no example for the new project yet, but there is a simple unit test for all JVM + native platforms. On Android, I'm running integration tests as part of the device check.


stevensJourney (approved):
The build process, while complex, is very impressive. This looks good to me :)

PR #294, opened 2025-11-26
Update SKIE, Kotlin, Gradle and Java

simolus3:
I noticed that some jobs were timing out, so this increases the timeout from 20 minutes to 30.

Also, in the hope of newer version = faster build, this updates Kotlin, Gradle and Java to the latest stable versions.


stevensJourney (approved):
Happy with the changes :D

PR #295, opened 2025-11-27
Use qualified reference to sqlite_sequence

simolus3:
Similar to the port for Dart, this ports https://github.com/powersync-ja/powersync-js/pull/781 to the Kotlin SDK.


stevensJourney (approved):
thanks for the port!

PR #300, opened 2025-12-03
Attach prebuilt libraries to GitHub releases

simolus3:
This adds a step to the release process that will attach the static `libsqlite3.a` builds to that release. After the first release using this, I'll add build logic to download from that release instead rebuilding from scratch on developer machines.

Downloading these files instead of compiling them locally should speed up our local workflow. Especially since the SQLite3MultipleCiphers build for the JVM will also require additional toolchains and only work on macOS hosts, downloading would also make it easier to contribute to the SDK.

This also changes a few references from the `:core` project to `:common`. Since the Supabase, Room and SQLDelight projects only need the `:common` project, this will allow them to be used with the encryption variant of the SDK.


stevensJourney (dismissed):
🚀

PR #305, opened 2025-12-09
Fix high memory usage on apple devices

joshuabrink:
Add a custom Ktor Darwin HTTP engine to the repository with bounded channel backpressure to prevent out-of-memory crashes when processing large sync payloads on iOS/macOS.

## Problem
NSURLSession delivers data faster than it can be processed, causing unbounded buffering in the upstream Ktor Darwin engine's channel. This results in multi-GB memory spikes and OOM crashes on large sync operations.

## Solution
Implemented a custom Darwin client fork with a bounded channel (capacity: 64) and backpressure handling via `runBlocking`. This naturally throttles NSURLSession's data delivery rate to match processing speed, eliminating memory spikes.

## Changes
- Added `internal/ktor-client-darwin/` with modified `DarwinTaskHandler.kt`
- Implements bounded channel backpressure for large payload handling
- Maintains full compatibility with existing PowerSync Kotlin SDK code
- Reduces memory footprint during large syncs by preventing unbounded buffering

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -91,6 +91,7 @@ ktor-client-darwin = { module = "io.ktor:ktor-client-darwin", version.ref = "kto
 ktor-client-okhttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
 ktor-client-contentnegotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
 ktor-client-mock = { module = "io.ktor:ktor-client-mock", version.ref = "ktor" }
+ktor-network-tls = { module = "io.ktor:ktor-network-tls", version.ref = "ktor" }
 ktor-serialization-json = { module = "io.ktor:ktor-serialization-kotlinx-json", version.ref = "ktor" }
 kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "serialization" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "coroutines" }
--- a/internal/ktor-client-darwin/build.gradle.kts
+++ b/internal/ktor-client-darwin/build.gradle.kts
@@ -32,6 +32,7 @@ kotlin {
             kotlin.srcDir("darwin/src")
             dependencies {
                 api(libs.ktor.client.core)
+                api(libs.ktor.network.tls)
                 implementation(libs.kotlinx.coroutines.core)
             }
         }


rkistner (commented):
Just some general comments:
1. Since most of the code is a fork from the ktor-client-darwin library, we need to include a copy of their license.
2. There is quite a lot of code here from the fork. Is there an easy way to view the difference? And how would we keep it up to date with the upstream version in the future? (Can we get the fix merged upstream?)
3. The bounded channel has a capacity of "64". What does this mean in practice? Does this translate to a practical limit in MB?

internal/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/issue.md
```diff
@@ -0,0 +1,45 @@
...
+## Problem Summary
+
+Out of Memory (OOM) error occurring during PowerSync sync operations when using the Kotlin Multiplatform SDK from Swift.
+
+## Root Cause
+
+The Ktor Darwin engine's `toByteArray()` function is loading entire HTTP response bodies into memory before processing them. This happens in two steps:
+
+1. **NSURLSession accumulates all response chunks** into a single `NSData` object
+2. **Ktor converts the entire NSData to ByteArray** - allocating another full copy
```
simolus3: This isn't what actually happens - if it were the case, streaming sync would be completely broken. We still see individual lines being emitted by ktor, it's just that backpressure doesn't reach the server.

simolus3 (commented):
The biggest question is whether this actually works. I've explored a different approach that calls `suspend()` on the `NSURLSession` download task when the channel is full, and the framework implementation just completely ignored that. So it's not obvious to me whether blocking the receive thread is enough to implement backpressure (since most of the OS APIs are asynchronous).

Steven has created an [aggressive repro](https://github.com/stevensJourney/ktor-memory-test) for this, can you check whether the client fork fixes that? You can start the server, then run the iOS app in a simulator. At least in my attempts, memory usage was growing unbounded for about 40% of all app starts (with my approach, it looks like there may be a race condition in the underlying HTTP client implementation from the OS).

joshuabrink:
> Steven has created an [aggressive repro](https://github.com/stevensJourney/ktor-memory-test) for this, can you check whether the client fork fixes that? You can start the server, then run the iOS app in a simulator. At least in my attempts, memory usage was growing unbounded for about 40% of all app starts (with my approach, it looks like there may be a race condition in the underlying HTTP client implementation from the OS).

Using a channel capacity of 1:
- Tested it on Stevens repo, we see stable memory usage ~80 MB (ran it for about 3 minutes and no change) 
- Also tested the using the Swift SDK directly (using the reproduction of the reported issue) and we see about ~700 MB stable memory usage during initial sync (it gets up to 900 MB intermittently but soon drops back down). Most of the additional memory seems to be PowerSync overhead:
```
 433.27 MB      74.2%	65420	 	                                                                              kfun:com.powersync.sqlite.Statement#step(){}kotlin.Boolean
 433.27 MB      74.2%	65420	 	                                                                               sqlite3_step
 433.27 MB      74.2%	65420	 	                                                                                sqlite3VdbeExec
 427.04 MB      73.1%	50405	 	                                                                                 powersync_core::operations_vtab::update::h057c31bbad73a7d0
 427.04 MB      73.1%	50405	 	                                                                                  powersync_core::operations::insert_operation::h5aafae70b4795e73
 427.04 MB      73.1%	50405	 	                                                                                   powersync_core::sync::operations::insert_bucket_operations::h2edc58eeb588e78b
 393.53 MB      67.3%	29768	 	                                                                                    sqlite3_capi::capi::step::h009b6dd0ba7a12c7
 393.53 MB      67.3%	29768	 	                                                                                     sqlite3_step
 393.53 MB      67.3%	29768	 	                                                                                      sqlite3VdbeExec
 348.84 MB      59.7%	3721	 	                                                                                       btreeBeginTrans
 348.84 MB      59.7%	3721	 	                                                                                        getPageNormal
 348.84 MB      59.7%	3721	 	                                                                                         pcache1FetchStage2
 348.84 MB      59.7%	3721	 	                                                                                          sqlite3Malloc
```

PR #315, opened 2026-01-29
Build `:internal:PowerSyncKotlin` for tvOS

simolus3:
While we've been building and testing tvOS builds of the Kotlin SDK for a while, I seemingly forgot to include tvOS as a supported target for the Swift SDK helper framework 🤦.

This adds the missing targets, allowing us to support tvOS in our Swift SDK (I ran unit tests with a local framework build).


stevensJourney (approved):
:shipit:

PR #318, opened 2026-02-05
Upgrade to Android Gradle Plugin version 9

simolus3:
This upgrades to AGP version 9.0. This is a major release of the plugin with a few breaking changes that affect our build.

With AGP version 9.0, a Kotlin plugin is added to all Android applications and libraries automatically. So we don't have to add the Kotlin Android plugin to applications anymore.

For multiplatform libraries, we have to keep using the multiplatfor Kotlin plugin, but have to replace the normal Android library Gradle plugin with a plugin variant optimized for libraries. This changes a few things:

1. The top-level `android {}` block to configure variants, namespaces and toolchain options is gone. Instead, we add a `androidLibrary {}` block as a KMP target (just like we do for the JVM and other native targets). This is actually pretty nice since it means we can share common properties like the target and build SDK version across projects.
2. It is no longer possible to add separate debug and release libraries though, there's just a [single variant](https://developer.android.com/kotlin/multiplatform/plugin#features). This affects our ability to statically detect whether the app is in debug mode. I've added a changelog entry explaining that users should manually configure the logger based on their `BuildConfig` now.
3. Also, it is no longer possible to add native builds to KMP libraries for Android. The [suggested workaround](https://developer.android.com/kotlin/multiplatform/plugin#unsupported-features-and-workarounds) is to add a separate, pure-Android library with the native build and depend on that in the KMP project. I've done that for the encryption library using JNI. At least the separate project lets us configure Android device tests more easily.

In my testing, I noticed some issues with the atomicfu Gradle plugin trying to access tasks that no longer exist. It seems like we can use ktor utilities for that instead, so I've just removed the plugin.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -139,7 +139,7 @@ kmp-lifecycle-compose = { module = "org.jetbrains.androidx.lifecycle:lifecycle-v
 
 [plugins]
 android-application = { id = "com.android.application", version.ref = "android-gradle-plugin" }
-#android-library = { id = "com.android.library", version.ref = "android-gradle-plugin" }
+android-library = { id = "com.android.library", version.ref = "android-gradle-plugin" }
 android-multiplatform-library = { id = "com.android.kotlin.multiplatform.library", version.ref = "android-gradle-plugin" }
 jetbrainsCompose = { id = "org.jetbrains.compose", version.ref = "compose" }
 compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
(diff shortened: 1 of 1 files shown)


common/src/androidMain/kotlin/BuildConfig.kt
```diff
@@ -1,5 +1,5 @@
 @Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
 internal actual object BuildConfig {
     actual val isDebug: Boolean
-        get() = com.powersync.common.BuildConfig.DEBUG
+        get() = false
```
simolus3: We no longer have a `BuildConfig` in this library since we can only generate a single variant.

We could inspect [the application's debug status](https://stackoverflow.com/a/23844693) at runtime, but that requires access to a `Context`. Since this is only used to infer a suitable default severity for the logger, I've added a changelog entry explaining the change.

Radiokot:
Hey guys, you released this as a patch version, yet it's a breaking update – it requires Gradle and compile SDK updates in an app and all the hell associated with it

PR #342, opened 2026-04-14
Remove Kotlin client implementation

simolus3:
Following the [similar Dart PR](https://github.com/powersync-ja/powersync.dart/pull/395), this removes the Kotlin sync client implementation (making the Rust client the only option).

This moves `SyncLine` classes into test sources (so that we can continue to use them in tests mocking the PowerSync service). All other `@LegacySyncImplementation` classes and methods have been removed.

I think we can get away with this in a minor release. While it is technically a breaking change, we've clearly communicated that using the non-default client implementation is not recommended.


rkistner (approved):
I did not check the code changes in detail, but the description makes sense: I'm happy for us to remove the old sync implementation in a minor release.

PR #343, opened 2026-04-16
Avoid reconnect delay on token prefetch

simolus3:
After pre-fetching a token (a process we start when the service indicates the current JWT is about to expire), we used to send a `stop` command to `powersync_control`. This causes us to stop the iteration and await a regular retry delay, which defeats the purpose of pre-fetching tokens in the first place.

The simple fix is to use the `refreshed_token` command instead.


common/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -558,4 +558,4 @@ internal class PowerSyncDatabaseImpl(
     }
 }
 
-internal object DisconnectRequestedException : CancellationException("disconnect() called")
+internal class DisconnectRequestedException : CancellationException("disconnect() called")
```
simolus3, line 561: This is an unrelated fix, exceptions shouldn't be singletons since the JVM installs a stack trace every time they're thrown.

stevensJourney (approved):

PR #344, opened 2026-04-28
Link core extension statically

### LINKED ISSUE #339: iOS: dyld crash after updating to 1.11.2 — powersync-sqlite-core.framework not embedded
## Description

After updating from PowerSync Kotlin **1.11.1** to **1.11.2** (and `powersync-sqlite-core-swift` from **0.4.11** to **0.4.12**), the iOS app crashes at launch with a `dyld` error:

```
dyld: Library not loaded: @rpath/powersync-sqlite-core.framework/powersync-sqlite-core
  Referenced from: NotPink.app/NotPink.debug.dylib
  Reason: tried: '…/NotPink.app/Frameworks/powersync-sqlite-core.framework/powersync-sqlite-core' (no such file)
```

The KMP shared framework (`NotPink.debug.dylib`) dynamically links `powersync-sqlite-core.framework` via `@rpath`, but the framework is not present in the app bundle's `Frameworks/` directory.

**Reverting to 1.11.1 + 0.4.11 resolves the issue.**

## Environment

- **PowerSync Kotlin SDK**: 1.11.2 (was 1.11.1)
- **powersync-sqlite-core-swift**: 0.4.12 (was 0.4.11)
- **Kotlin**: 2.1.x
- **Xcode**: 26.0
- **iOS target**: 26.0
- **Project setup**: Kotlin Multiplatform with Tuist-generated Xcode project

## Setup

The iOS app declares `powersync-sqlite-core-swift` as an SPM dependency and references it in the app target. The KMP shared framework is built via `./gradlew :shared:embedAndSignAppleFrameworkForXcode`.

With **1.11.1 / 0.4.11**, the app launches fine — `powersync-sqlite-core` does not appear as a dynamic dependency of the shared framework.

With **1.11.2 / 0.4.12**, the shared framework (`NotPink.debug.dylib`) gains a dynamic link to `@rpath/powersync-sqlite-core.framework/powersync-sqlite-core`, which is not automatically embedded in the app bundle by the build system.

## Expected behavior

The PowerSync Kotlin SDK should either:
1. Statically link `powersync-sqlite-core` into the shared framework (as it appeared to do in 1.11.1), or
2. Document that `powersync-sqlite-core.framework` must be manually embedded in the iOS app bundle, with setup instructions for common project configurations (SPM, CocoaPods, Tuist, etc.)

## Steps to reproduce

1. Create a KMP project using `com.powersync:core:1.11.2`
2. Add `powersync-sqlite-core-swift` 0.4.12 as an SPM dependency in the iOS app
3. Build and run on iOS simulator
4. App crashes at launch with the dyld error above

## Workaround

Pin to `com.powersync:core:1.11.1` and `powersync-sqlite-core-swift:0.4.11`.

simolus3:
Currently, we don't link the core extension at all on native platforms. Instead:

1. We ask users to [manually depend on it](https://docs.powersync.com/client-sdks/reference/kotlin#direct-integration) and then load it as a path.
2. For our tests, we download desktop libraries and patch the link task of tests to include it.

Following this setup is annoying and error-prone for users, since they have to manage this dependency outside of Gradle and remember to update it every time they upgrade the Kotlin SDK.

This fixes the issue by linking the core extension statically: The `:core` package uses a cinterop definition with a downloaded static library attached to core extension releases. This embeds the core extension with the built klib, and users don't have to rely on external dependenies anymore. Additionally, it simplifies our loading setup: Previously, we had to use `sqlite3_load_extension` with a resolved framework path. With static linking, we can simply call `sqlite3_auto_extension` with the `sqlite3_powersync_init` function pointer to load the extension.

I've tested the demo app in a simulator and a local framework build for the Swift SDK (that currently causes additional test failures there because Kotlin doesn't support the old sync client anymore, but we're able to load the extension).

Closes https://github.com/powersync-ja/powersync-kotlin/issues/339.


core/build.gradle.kts
```diff
@@ -62,7 +62,6 @@ kotlin {
 
         androidMain {
             dependencies {
-                api(libs.powersync.sqlite.core.android)
```
simolus3, line 65: We already depend on this in `:common`, this clarifies that the `:common` package is solely responsible for loading the core extension on all platforms.

stevensJourney (approved):
The changes here look good to me. This makes the setup steps much simpler for users. I can't think of any negative repercussions for these changes.

PR #345, opened 2026-05-06
Remove `:internal:PowerSyncKotlin`, prepare 1.12.0 release

simolus3:
Since the Swift SDK no longer relies on it, this removes the internal `:PowerSyncKotlin` XCFramework, SKIE dependencies and Objective C interop annotations. In the `:common` project, it removes the split between public and serializable schema classes, slightly simplifying our setup there (we've only split those classes to avoid serializable symbols leaking into the XCFramework).

This also updates a few build and Github actions dependencies and prepares the 1.12.0 release.


stevensJourney (approved):
🚀

PR #347, opened 2026-05-12
Update README to reflect beta status

benitav:
Now that raw table APIs are stable, this can advance to beta.


benitav:
Superseded by https://github.com/powersync-ja/powersync-kotlin/pull/346

PR #348, opened 2026-05-13
Add isInitializing to SyncStatusData to distinguish loading state from empty streams

Sagbyy:
## Summary

`SyncStatusData.syncStreams` returned an empty list `[]` in two distinct situations that were previously indistinguishable:

- **During initial database setup** — before the core extension emits any sync status, `internalSubscriptions` defaulted to `emptyList()`, making it impossible to know whether the database was still opening or whether there were genuinely no active streams.
- **After sync, with no active streams** — a legitimate state where the database is fully initialized but no streams are being tracked.

This meant consumers could not reliably react to the initial loading state without waiting for `downloading`, `connecting`, or other flags that only become available after a connection attempt.

## Solution

Added a boolean attribute `isInitializing` to `SyncStatusData`:

- Defaults to `true` in `SyncStatusDataContainer`, representing the period before the first core status update.
- Set to `false` in `applyCoreChanges()`, called by `resolveOfflineSyncStatus()` during database initialization.
- No nullable types introduced — `syncStreams` continues to return a non-null list at all times.

```kotlin
  when {
      status.isInitializing       -> // DB still opening, state unknown
      status.syncStreams.isEmpty() -> // initialized, no active streams
      else                        -> // initialized, streams available
  }
```

## Tests

Added two integration tests in SyncStreamTest:
- isInitializing is true before database init and false after — verifies the true → false transition
- isInitializing remains false after connecting — verifies connecting does not reset the flag

### DIFF
--- a/common/src/commonIntegrationTest/kotlin/com/powersync/sync/SyncStreamTest.kt
+++ b/common/src/commonIntegrationTest/kotlin/com/powersync/sync/SyncStreamTest.kt
@@ -28,6 +28,33 @@ import kotlin.test.Test
 @OptIn(ExperimentalPowerSyncAPI::class)
 class SyncStreamTest : AbstractSyncTest(true) {
     @Test
+    fun `isInitializing is true before database init and false after`() =
+        databaseTest(createInitialDatabase = false) {
+            database = openDatabase()
+            // Init job is queued but not yet started — isInitializing must be true
+            database.currentStatus.isInitializing shouldBe true
+
+            // readLock suspends until initialization completes (resolveOfflineSyncStatus is called)
+            database.readLock { }
+            database.currentStatus.isInitializing shouldBe false
+        }
+
+    @Test
+    fun `isInitializing remains false after connecting`() =
+        databaseTest {
+            database.currentStatus.isInitializing shouldBe false
+
+            database.connect(connector, options = getOptions())
+            turbineScope {
+                val turbine = database.currentStatus.asFlow().testIn(this)
+                turbine.waitFor { it.connected && !it.downloading }
+
+                database.currentStatus.isInitializing shouldBe false
+                turbine.cancelAndIgnoreRemainingEvents()
+            }
+        }
+
+    @Test
     fun `can disable default streams`() =
         databaseTest {
             database.connect(
--- a/common/src/commonIntegrationTest/kotlin/com/powersync/sync/SyncStreamTest.kt
+++ b/common/src/commonIntegrationTest/kotlin/com/powersync/sync/SyncStreamTest.kt
@@ -28,28 +28,28 @@ import kotlin.test.Test
 @OptIn(ExperimentalPowerSyncAPI::class)
 class SyncStreamTest : AbstractSyncTest(true) {
     @Test
-    fun `isInitializing is true before database init and false after`() =
+    fun `syncStreams is null before database init and non-null after`() =
         databaseTest(createInitialDatabase = false) {
             database = openDatabase()
-            // Init job is queued but not yet started — isInitializing must be true
-            database.currentStatus.isInitializing shouldBe true
+            // Init job is queued but not yet started — syncStreams must be null
+            database.currentStatus.syncStreams shouldBe null
 
             // readLock suspends until initialization completes (resolveOfflineSyncStatus is called)
             database.readLock { }
-            database.currentStatus.isInitializing shouldBe false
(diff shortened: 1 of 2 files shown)


simolus3 (commented):
I like the idea of adding an explicit initialization state to `SyncStatus`, but it's something we'd have to do across all our SDKs.

I see that the `syncStreams` getter is already nullable, and documented to return `null` during initialization. So it sounds like a possible fix without API changes might be to make `SyncStatusDataContainer.internalSubscriptions` null initially (the `emptyList()` default feels wrong). Then, you can use `null` vs `emptyList` as a distinguishing factor.

Sagbyy:
> I like the idea of adding an explicit initialization state to `SyncStatus`, but it's something we'd have to do across all our SDKs.
> 
> I see that the `syncStreams` getter is already nullable, and documented to return `null` during initialization. So it sounds like a possible fix without API changes might be to make `SyncStatusDataContainer.internalSubscriptions` null initially (the `emptyList()` default feels wrong). Then, you can use `null` vs `emptyList` as a distinguishing factor.

Thanks for the suggestion, I've applied the changes `internalSubscriptions` now defaults to `null` so `syncStreams` correctly returns `null` during initialization and an empty list once the database is ready.

I totally understand that adding `isInitializing` across all SDKs is a significant constraint, so I'm fine dropping that approach.

That said, I'd like to raise a related but deeper issue that this PR doesn't solve. When a resource change occurs (switching organization or user), the sequence is:
1. Table is cleared → `syncStreams` / data queries return `[]`
2. Brief gap where `hasSynced = true`, `connected = true`, `downloading = false`
3. Re-sync starts → `downloading = true`

During step 2, the status looks "ready" but the data is actually empty and stale. There's currently no way to distinguish this from a legitimate "synced and empty" state. `connecting || downloading` doesn't fully cover it either since there's a window where both are false.

Is there a workaround on our side without SDK changes ? Or would it be worth considering a `dataStale: Boolean` flag (or similar) in `SyncStatusData` that gets set to `true` when data is cleared before a re-sync ? This feels like it would benefit all SDKs anyway, and could be added progressively without breaking changes.

simolus3 (approved):

simolus3:
> When a resource change occurs (switching organization or user), the sequence is:

Does this involve a `disconnectAndClear`call? I agree that resetting `hasSynced` should ideally happen along with the actual clear. We seem to disconnect before clearing the database though, so I wonder how that state leaked 🤔

Sagbyy:
> > When a resource change occurs (switching organization or user), the sequence is:
> 
> Does this involve a `disconnectAndClear`call? I agree that resetting `hasSynced` should ideally happen along with the actual clear. We seem to disconnect before clearing the database though, so I wonder how that state leaked 🤔

In our case, `disconnectAndClear` is only called on logout it's not involved in the scenario I described.

What we observe is triggered purely by calling `connect()` with new parameters (when the user switches to a different resource). Right after `connect()` is called, there's a brief window where `connected=true`, `hasSynced=true`, `connecting=false`, `downloading=false` before `connecting=true` kicks in. During that window, `watch()` queries return `[]` and our status check (`connecting || downloading`) is `false`, so the UI incorrectly treats it as a "synced and empty" state.

We're working around it by debouncing the transition to `false` on the status signal if `connecting=true` before the delay, we cancel and stay in loading. It works, but it's a heuristic.

A `dataStale: Boolean` flag set between `connect()` and the first `downloading=true` would cleanly solve this without any debounce.

simolus3:
> What we observe is triggered purely by calling `connect()` with new parameters (when the user switches to a different resource).

I see.

> Right after `connect()` is called, there's a brief window where `connected=true`, `hasSynced=true`, `connecting=false`, `downloading=false`

Calling `connect()` should internally disconnect first. Maybe this just gets lost in some debouncing logic, but the steps should be `(connected: true, connecting: false, downloading: false)` (idle) -> `connect()` is called -> `(connected: false)` -> `(connecting: true)` -> `(connected: true, connecting: false, downloading: false)` -> `(..., downloading: true)`.

> During that window, `watch()` queries return `[]` and our status check (`connecting || downloading`) is `false`, so the UI incorrectly treats it as a "synced and empty" state.

I assume that's because the queries watch for the new resource, right? But couldn't you check for the `hasSynced` status on the stream subscription for the resource (assuming a resource switch also adds a new subscription?) to check whether the new data has synced.

The global `hasSynced` field isn't all that useful with Sync Streams, but `!connecting && !downloading` does not imply that the database has synced (it could also be genuinely offline).

PR #352, opened 2026-05-25
fix: bump supabase-kt to 3.6.0 to fix binary incompatibility in connector-supabase

### LINKED ISSUE #351: connector-supabase: binary incompatibility with supabase-kt 3.6.0 — NoSuchMethodError on all CRUD uploads
## Description

When using `com.powersync:connector-supabase:1.11.2` (or 1.12.0 — `main` has the same pin) with `supabase-kt 3.6.0`, **all CRUD uploads fail at runtime** with `NoSuchMethodError`. No rows can be created, updated, or deleted in Supabase. The error is silently retried forever by the sync loop (`StreamingSync.kt`), so the app doesn't crash but data never reaches the backend.

```
PowerSyncException: No direct method <init>(Lio/github/jan/supabase/postgrest/PropertyConversionMethod;)V
in class Lio/github/jan/supabase/postgrest/query/request/UpsertRequestBuilder;
```

## Root cause

The postgrest functions `upsert()`, `update()`, and `delete()` on `PostgrestQueryBuilder` are declared as **`inline`** in supabase-kt. When `connector-supabase` is compiled against supabase-kt 3.2.2, the Kotlin compiler inlines the function bodies — including their internal constructor calls — directly into `SupabaseConnector`'s bytecode.

In supabase-kt 3.6.0 ([PR #1264](https://github.com/supabase-community/supabase-kt/pull/1264)), all request builder constructors gained a `defaultSchema: String` parameter:

| Class | 3.2.2–3.5.0 | 3.6.0 |
|---|---|---|
| `UpsertRequestBuilder` | `(PropertyConversionMethod)` | `(String, PropertyConversionMethod)` |
| `InsertRequestBuilder` | `(PropertyConversionMethod)` | `(String, PropertyConversionMethod)` |
| `PostgrestRequestBuilder` | `(PropertyConversionMethod)` | `(String, PropertyConversionMethod)` |
| `UpdateRequestBuilder` | _(didn't exist)_ | `(String, PropertyConversionMethod)` |
| `DeleteRequestBuilder` | _(didn't exist)_ | `(String, PropertyConversionMethod)` |

The stale inlined bytecode in the connector JAR calls the old single-arg constructors, which no longer exist at runtime → `NoSuchMethodError`.

**All three CRUD operations are affected** (PUT/upsert, PATCH/update, DELETE/delete), not just upsert — the error just surfaces on whichever operation fires first.

## Why it's silent

1. The `NoSuchMethodError` (extends `Error`, not `Exception`) bypasses the `catch (e: Exception)` in `SupabaseConnector.uploadData()`, so `handleError()` is never invoked.
2. `runWrapped` in `db/Functions.kt` catches `Throwable` and wraps it in `PowerSyncException` (which extends `Exception`).
3. `StreamingSync.kt`'s upload loop catches `Exception`, logs via Kermit (`"Error uploading crud: …"`), delays, and retries — silently, forever.

No crash, no callback to the consumer, no way to detect this from the app side except by observing that rows stop arriving in Supabase.

## Workaround

Override `uploadCrudEntry` in a subclass of `SupabaseConnector` with the same logic. The override compiles against the app's actual supabase-kt version, so the `inline` functions produce correct bytecode:

```kotlin
override suspend fun uploadCrudEntry(entry: CrudEntry) {
    val table = supabaseClient.from(entry.table)
    when (entry.op) {
        UpdateType.PUT -> {
            val data = buildMap {
                put("id", JsonPrimitive(entry.id))
                entry.opData?.jsonValues?.let { putAll(it) }
            }
            table.upsert(data)
        }
        UpdateType.PATCH -> {
            table.update(entry.opData!!.jsonValues) {
                filter { eq("id", entry.id) }
            }
        }
        UpdateType.DELETE -> {
            table.delete {
                filter { eq("id", entry.id) }
            }
        }
    }
}
```

## Suggested fixes

1. **Short-term:** Recompile `connector-supabase` against supabase-kt 3.6.0 (or at least 3.5.0 + 3.6.0 tested).
2. **Short-term:** Add a Gradle dependency constraint (e.g. `api("io.github.jan-tennert.supabase:postgrest-kt") { version { strictly("[3.0.0, 3.5.0]") } }`) so consumers get a build-time error instead of a silent runtime failure when they upgrade supabase beyond the tested range.
3. **Long-term:** Avoid depending on the internal implementation details of `inline` functions from supabase-kt. Since the connector's `uploadCrudEntry` just calls `table.upsert/update/delete`, consider making `uploadCrudEntry` non-inline-dependent (e.g. use the lower-level HTTP API or request supabase-kt to stabilize these constructors).

## Environment

- `com.powersync:connector-supabase`: 1.11.2 (also reproduced concept on 1.12.0 — same supabase pin on `main`)
- `io.github.jan-tennert.supabase:bom`: 3.6.0
- Platform: Android (Dalvik/ART)
- Kotlin: 2.3.21

Shamyyoun:
## Summary

Bumps supabase-kt from 3.2.2 to 3.6.0 in `gradle/libs.versions.toml`.

The postgrest functions `upsert()`, `update()`, and `delete()` on `PostgrestQueryBuilder` are `inline` in supabase-kt. When `connector-supabase` is compiled against 3.2.2, the Kotlin compiler inlines their internal constructor calls into the connector's bytecode. In supabase-kt 3.6.0, all request builder constructors gained a `defaultSchema: String` parameter, so any consumer who upgrades supabase-kt past 3.5.0 hits `NoSuchMethodError` on every CRUD upload at runtime — silently retried forever by the sync loop with no crash and no callback.

This is a one-line version bump. The call-site API (`table.upsert()`, `table.update()`, `table.delete()`) is unchanged in 3.6.0, so connector source code compiles without modifications — the fix is simply recompiling against the new constructors.

Fixes #351

### DIFF
--- a/demos/supabase-todolist/androidBackgroundSync/src/main/java/com/powersync/demo/backgroundsync/SyncService.kt
+++ b/demos/supabase-todolist/androidBackgroundSync/src/main/java/com/powersync/demo/backgroundsync/SyncService.kt
@@ -1,3 +1,5 @@
+@file:OptIn(ExperimentalAtomicApi::class)
+
 package com.powersync.demo.backgroundsync
 
 import android.app.Notification
@@ -14,10 +16,11 @@ import com.powersync.PowerSyncDatabase
 import com.powersync.connector.supabase.SupabaseConnector
 import com.powersync.sync.SyncStatusData
 import io.github.jan.supabase.auth.status.SessionStatus
-import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.launch
 import org.koin.android.ext.android.inject
+import kotlin.concurrent.atomics.AtomicBoolean
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
 
 class SyncService: LifecycleService() {
 
@@ -92,7 +95,7 @@ class SyncService: LifecycleService() {
 
     override fun onDestroy() {
         if (holdsServiceLock) {
-            SERVICE_RUNNING.value = false
+            SERVICE_RUNNING.store(false)
         }
 
         super.onDestroy()
@@ -118,6 +121,6 @@ class SyncService: LifecycleService() {
 
     private companion object {
         val CHANNEL_ID = "background_sync"
-        val SERVICE_RUNNING = atomic(false)
+        val SERVICE_RUNNING = AtomicBoolean(false)
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix sync status to make `syncStreams` return null when the database is initializing.
 - Fix errors when subscribing to Sync Streams with object or array parameters.
+- Support versions 3.6.0 of the Supabase client libraries.
 
 ## 1.12.0
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -23,7 +23,7 @@ turbine = "1.2.1"
 kotest = "5.9.1" # we can't upgrade to 6.x because that requires Java 11 or above (we need Java 8 support)
 
 stately = "2.1.0"
-supabase = "3.2.2"
+supabase = "3.6.0"
 junit = "4.13.2"
 
 compose = "1.9.3" # This is for the multiplatform compose


demos/supabase-todolist/androidBackgroundSync/src/main/java/com/powersync/demo/backgroundsync/SyncService.kt
```diff
@@ -1,3 +1,5 @@
+@file:OptIn(ExperimentalAtomicApi::class)
```
simolus3, line 1: I've migrated this to stdlib atomics, atomicfu is no longer in the classpath since the Supabase client does't use it anymore.

simolus3 (approved):
Thanks for the contribution!

stevensJourney (approved):

PR #353, opened 2026-05-28
Update dependencies, remove macOS x64 support

simolus3:
This updates dependencies and removes support for x64 macOS platforms for Kotlin/native (we still support Intel macs on the JVM).

## Product visibility

This removes support for x64 macOS platforms and simulators (iOS, watchOS and macOS), which is a breaking change. This change is motivated by platform changes:

- Kotlin is [deprecating these targets](https://youtrack.jetbrains.com/issue/KT-78660/meta-Deprecation-of-the-macosX64-target), and will remove them in a future release. (actually, they've paused the deprecation and will revisit this after WWDC)
- AndroidX multiplatform has already removed support for them (https://github.com/androidx/androidx/commit/7cb9a4cc835cbdd98b53c1bb7bc3e0f59c6510fd), and we use AndroidX dependencies for SQLite interfaces. So keeping them around means we can't update dependencies.

In practice, I don't think this is a severe removal:

- Compose Multiplatform apps are based on the JVM for Desktop targets, so they continue to work.
- On iOS, watchOS and tvOS, only simulators are affected.

So I'm not sure there's much of a use case for a native macOS target, apart from maybe running tests?


stevensJourney (approved):
This feels like the logical move to take given the support is being dropped by our dependencies. 

cc @LucDeCaf  for context with documenting supported platforms.

PR #357, opened 2026-06-06
Fix read-pool lease leak that deadlocks updateSchema (#356)

### LINKED ISSUE #356: PowerSyncDatabase.updateSchema() deadlocks on same-schema call before connect()
PowerSyncDatabase.updateSchema(schema) hangs forever (no return, no error, no status change) when:
1. The schema argument is the same Schema instance the database was constructed with, AND
2. connect(...) has not yet been called on the database.

Confirmed reproducible on PowerSync 1.11.2 and 1.12.0 (Android target, debug builds). Calling updateSchema with a different schema instance works fine and returns in ~25ms. Only the same-instance case hangs.

**Reproduction**
```
val schema = Schema(/* whatever */)
val db = PowerSyncDatabase(factory, schema, "test.db", scope)
// no db.connect(...) call

// hangs forever, never returns, no exception, no log:
db.updateSchema(schema)
```

**Impact**

This affects our app on every cold launch. Our auth flow eventually calls updateSchema(schema) in response to Supabase's Authenticated status, either immediately on launch or shortly after. In both cases the schema passed to updateSchema is the same instance the  database was constructed with, and the call never returns, blocking the rest of the auth/sync flow indefinitely.

At the time of the hung call, powerSyncDatabase.currentStatus reads:
```
SyncStatus(connected=false, connecting=false, downloading=false, uploading=false, lastSyncedAt=null, hasSynced=false, error=null)
```

The database is in its initial state, no connection ever attempted. The updateSchema coroutine suspends and never resumes. A withTimeoutOrNull(10.seconds) wrapper does successfully cancel it, so the suspension is at least cancellable.

**Expected behavior**

updateSchema() with the same schema as currently installed should be a no-op that returns immediately, or complete in bounded time without hanging.

**Workaround**

Track the installed schema in app code and short-circuit redundant calls:

```
private var installedSchema: Schema? = null  // initialized to whatever was passed at construction

private suspend fun ensureSchema(target: Schema) {
  if (installedSchema === target) return
  db.updateSchema(target)
  installedSchema = target
}
```

This avoids the deadlock entirely. Legitimate schema changes still complete in ~25ms.

**Environment**
- powersync-kotlin: 1.11.2 (also reproduced on 1.12.0)
- Android target
- Kotlin 2.x, kotlinx-coroutines 1.10.x

marioortizmanero:
## What

`ReadPool`'s connection-lease handoff channel (`available`) lacked an `onUndeliveredElement` handler. When a `read {}` (or the `withAllConnections` acquisition loop) is **cancelled during the channel rendezvous** — after a worker `send`s `(connection, done)` but before the receiver delivers it into its `try/finally` — the element is silently dropped. `done.complete(Unit)` is never called, so that worker parks in `done.await()` **forever** and the connection is permanently lost from the pool.

Once any lease leaks, `ReadPool.withAllConnections { repeat(size) { available.receive() } }` can never acquire all connections and **hangs forever**. `updateSchema()` reaches the pool through exactly this path:

```
PowerSyncDatabaseImpl.updateSchema
  → mutex.withLock
    → InternalDatabaseImpl.updateSchema
      → InternalConnectionPool.withAllConnections
        → ReadPool.withAllConnections → repeat(size){ available.receive() }
```

so a single leaked read lease wedges `updateSchema` indefinitely while it holds the database mutex.

## Why #356 looked schema-related

It isn't about passing the same `Schema` instance or about `connect()`. The reporter's app calls `updateSchema` on cold start while concurrent `watch()` queries are being cancelled (auth churn over a dead/slow network) — that cancellation is what leaks the lease. A standalone `updateSchema` with no concurrent reads (the minimal repro) never leaks, which is why it passed.

## Fix

Add `onUndeliveredElement` to the `available` channel:

```kotlin
onUndeliveredElement = { (_, done) -> done.complete(Unit) }
```

`onUndeliveredElement` fires exactly when an element was handed to the channel but not received (cancelled receiver, or `close()`/`cancel()` with an in-flight element). Completing `done` returns the connection to circulation. `CompletableDeferred.complete()` is idempotent and non-suspending, so there's no double-completion hazard with the normal `finally` paths. This covers both leak windows — `read()` and the `withAllConnections` `add(receive())` gap — and the close path.

## How we found this

A thread dump was useless — every thread was idle/parked with no SQLite or native frame (a suspended-coroutine deadlock, not a blocked thread). Coroutine debug probes are inert on Android/ART, so we reflected into the live pool from a debug build at the moment of the hang: all 5 read workers alive, DB mutex held, `writeLockMutex` still free, but one connection checked out and never returned — i.e. `withAllConnections` stuck on its final `available.receive()` after a leaked read lease. That pointed straight at `ReadPool` and the missing `onUndeliveredElement`.

## Test

`ReadPoolLeaseLeakTest` hammers the read pool with concurrent reads cancelled at varied points, then asserts `updateSchema` completes under a timeout. It runs on real dispatchers (the race is hidden by the virtual-time test dispatcher).

- **Before the fix:** fails with `TimeoutCancellationException` from `updateSchema`.
- **After the fix:** passes in a few ms.

Full `:common:jvmTest` suite: 130 tests, 0 failures.

Fixes #356.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - __Breaking change__: Aligning with Kotlin and Androidx multiplatform libraries, this 
   release removes support for `x86_64` Apple targets (`iosX64()`, `macosX64()`, `tvosX64()` and `watchosX64()`).
+- Fix a deadlock where a read query cancelled at the wrong moment would permanently leak a
+  connection from the read pool, eventually causing `updateSchema()` (and other operations that
+  acquire all connections) to hang. This is the underlying cause of #356.
 - Fix sync status to make `syncStreams` return null when the database is initializing.
 - Fix errors when subscribing to Sync Streams with object or array parameters.
 - Support versions 3.6.0 of the Supabase client libraries.
--- a/common/src/commonIntegrationTest/kotlin/com/powersync/ReadPoolLeaseLeakTest.kt
+++ b/common/src/commonIntegrationTest/kotlin/com/powersync/ReadPoolLeaseLeakTest.kt
@@ -1,0 +1,95 @@
+package com.powersync
+
+import co.touchlab.kermit.Logger
+import co.touchlab.kermit.Severity
+import co.touchlab.kermit.StaticConfig
+import co.touchlab.kermit.platformLogWriter
+import com.powersync.db.schema.Schema
+import com.powersync.test.factory
+import com.powersync.test.getTempDir
+import com.powersync.testutils.UserRow
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.joinAll
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.test.Test
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Regression test for https://github.com/powersync-ja/powersync-kotlin/issues/356
+ *
+ * A `read {}` (or the `withAllConnections` acquisition loop) cancelled during the channel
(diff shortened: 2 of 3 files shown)


simolus3 (approved):
Good catch, thanks for the fix!

PR #359, opened 2026-06-22
Refactor publishing workflow, reduce file count

simolus3:
To be able to understand what we're uploading to Maven Central better, this restructures the publishing workflow. Currently, publishing works by:

1. In each package we publish, writing a release to a `build/sonatypeLocal` maven repository.
2. In each package we publish, invoking a custom Gradle task zipping that local repository and uploading it via the central publishing API.

This keeps the first step, but refactors the second to first aggregate all packages being published in the top-level `build/sonatypeBundle` directory. This allows us to analyze the bundle from a single directory. Also, it allows us to replace custom uploading code and Gradle tasks with a plain `curl` invocation.

Additionally, this fixes the build configuration to avoid dokka output being published as a `-javadoc.jar`. Those take up a lot of space, and aren't useful since we publish them separately. Also, skips sha512 hashes (only keeping md5, sha1 and sha256 around) and hashes on signatures (which aren't required), reducing the total amount of files.

I also had Claude generate an analysis script to break down the total file size. This is the output from a test run in CI:

```
Total: 40.6 MB  (2435 files)

By file type:
  Type                       Size   Count
  -------------------- ----------  ------
  .klib                   24.3 MB      87
  .jar                     9.5 MB      70
  .aar                     4.7 MB       8
  sources jar              1.3 MB      78
  .module                502.2 KB      78
  .pom                   203.0 KB      78
  .asc signature          92.7 KB     487
  checksum                64.7 KB    1461
  other                   23.4 KB      10
  javadoc jar             19.9 KB      78
```

This is a noticable reduction in total file size (which was at roughly 100MB before this PR) and file count (was at 5k). However, it's still too much. We should discuss next steps offline.

_AI use disclosure_: The analysis script and some build changes were generated with Claude Code, but manually revised.


Manrich121 (approved):
LGTM and excluding sha512 hashes make sense

PR #367, opened 2026-08-10
Update dependencies

simolus3:
This updates dependencies and fixes a few deprecation warnings:

- Update Kotlin to 2.4.10, the latest version.
- Update to Android SDK version 37.
- Remove unused uuid library dependencies (Kotlin has a builtin uuid type now, also we're not using that library at all).
- For Compose Multiplatform dependencies, avoid deprecated imports from Gradle plugin in favor of manual dependencies.
- Remove dependency on external material icons dependency, the recommended approach is to [add icons as resources](https://kotlinlang.org/docs/multiplatform/compose-multiplatform-resources-usage.html#icons) instead.

I have tested the supabase demo on Android with these changes.


common/src/commonMain/kotlin/com/powersync/db/internal/ConnectionContext.kt
```diff
@@ -135,9 +135,7 @@ internal fun SQLiteStatement.bind(parameters: List<Any?>?) {
             }
 
             else -> {
-                if (parameter != null) {
```
simolus3, line 138: We handle a `null` case above, so this always evaluates to `true`.

Manrich121 (approved):

PR #371, opened 2026-08-19
New checkpoint request protocol

simolus3:
This adds opt-in support for the new checkpoint request endpoint, similar to [this JS PR](https://github.com/powersync-ja/powersync-js/pull/1073).

With the new endpoint, clients keep track of their local request state and reconcile it with the PowerSync service at the beginning of a sync iteration. For the most part, the core extension tells us what to do. The main change in the implementation is to delay checkpoint requests until that initial reconciliation is complete. We also need to periodically re-request checkpoints that haven't been applied yet as the service is allowed to forget about them.

AI use: Tests ported from the JavaScript SDK with Claude Code. I also used Claude to deflake a few tests.

### DIFF
--- a/common/src/commonNonWeb/kotlin/com/powersync/db/driver/InternalConnectionPool.kt
+++ b/common/src/commonNonWeb/kotlin/com/powersync/db/driver/InternalConnectionPool.kt
@@ -6,33 +6,29 @@ import com.powersync.ExperimentalPowerSyncAPI
 import com.powersync.PersistentConnectionFactory
 import com.powersync.utils.JsonUtil
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.IO
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharedFlow
-import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
+import kotlin.coroutines.CoroutineContext
 
 @OptIn(ExperimentalPowerSyncAPI::class)
 internal class InternalConnectionPool(
     private val factory: PersistentConnectionFactory,
-    private val scope: CoroutineScope,
+    scope: CoroutineScope,
     private val dbFilename: String,
     private val dbDirectory: String?,
     private val writeLockMutex: Mutex,
+    /**
+     * Database calls are synchronous/blocking, so we always run them on Dispatchers.IO instead of
+     * inheriting the caller-provided scope context. The provided scope is still used for
+     * lifecycle-bound pool coroutines like read workers and update emission.
+     */
+    private val dispatcher: CoroutineContext,
 ) : SQLiteConnectionPool {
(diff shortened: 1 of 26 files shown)


common/src/commonNonWeb/kotlin/com/powersync/db/driver/InternalConnectionPool.kt
```diff
@@ -6,34 +6,30 @@ import com.powersync.ExperimentalPowerSyncAPI
...
+    scope: CoroutineScope,
     private val dbFilename: String,
     private val dbDirectory: String?,
     private val writeLockMutex: Mutex,
+    /**
+     * Database calls are synchronous/blocking, so we always run them on Dispatchers.IO instead of
+     * inheriting the caller-provided scope context. The provided scope is still used for
+     * lifecycle-bound pool coroutines like read workers and update emission.
+     */
+    private val dispatcher: CoroutineContext,
```
simolus3, line 28: Not using `Dispatchers.IO` is required for some checkpoint request tests which deliberately race the upload and download loop to verify that requesting a checkpoint resumes a download iteration.

When an IO dispatcher is used, the test dispatcher can't observe that coroutine and advances its mocked timer. Consistently using the test dispatcher is a best practice anyways, now we only use the IO dispatcher for tests that deliberately test parallelism and slow queries across threads.

common/src/commonNonWeb/kotlin/com/powersync/db/driver/InternalConnectionPool.kt
```diff
@@ -55,18 +51,17 @@ internal class InternalConnectionPool(
...
-                        }
-                    }
+                }
+            } finally {
+                // When we've leased a write connection, we may have to update table update flows
+                // after users ran their custom statements. Reading updates is a SQL call, but it
+                // doesn't do IO so we can do it on the main dispatcher.
+                val updatedTables = writeConnection.readPendingUpdates()
+                if (updatedTables.isNotEmpty()) {
+                    updates.emit(updatedTables)
```
simolus3: There is no need to wrap this in a `scope.launch`: `tableUpdatesFlow` is only observed internally, and only in a way that doesn't block:

1. We have one listener in `connectInternal` to call `triggerCrudUpload()`, which is non-blocking.
2. Watched queries are debounced in `rawChangedTables`, our throttle extension there uses `buffer()` with a conflated channel to avoid blocking.

common/src/commonNonWeb/kotlin/com/powersync/db/driver/InternalConnectionPool.kt
```diff
@@ -53,23 +49,27 @@ internal class InternalConnectionPool(
...
+                        callback(RawConnectionLease(writeConnection))
+                    } finally {
+                        // When we've leased a write connection, we may have to update table update
+                        // flows after users ran their custom statements.
+                        updatedTables = writeConnection.readPendingUpdates()
                     }
                 }
             }
+        } finally {
+            // Emit updates on teh main context, and outside the write lock.
```
Manrich121:
```suggestion
            // Emit updates on the main context and outside the write lock.
```

Manrich121 (dismissed):

Manrich121 (approved):

PR #372, opened 2026-08-19
Sync client improvements

simolus3:
This ports some of the sync client improvements we've recently implemented on our other SDKs to the Kotlin SDK. The main goal here is to reduce technical debt, thereby making checkpoint requests simpler to implement.

There is one improvement per commit:

1. Restructure CRUD uploads: In the Kotlin SDK, crud uploads are triggered as jobs lauched in a coroutine scope, one per upload attempt. To avoid concurrent upload jobs, we keep an atomic reference to bail out early. All our other SDKs use a much cleaner approach: Let a continous asynchronous loop drive CRUD uploads, and trigger iterations through channels. This ports the loop [from JavaScript](https://github.com/powersync-ja/powersync-js/blob/463381c89ca3fa821e4e1a6e35792b90152f9086/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts#L232-L242) which is a neat simplification as we can use structured concurrency to scope uploads and downloads. No more `CoroutineScope`s being passed around! A behavioral change inherited from JavaScript is that we now start one upload iteration whenever we connect.
2. Instead of letting the sync client manage its own mutable sync status that is then copied into the database through an async job, just let the sync client mutate the status on the database directly. It's already atomic, stopping the copy means there's no way for the sync client to keep updating its status while the copy job is already cancelled.
3. Drive the sync status from core state: In our other SDKs, the publicly visible sync status is derived from what the core extension gives us, download/upload errors and a `bool isUploading` flag. This ports that approach to Kotlin, dropping some code in `SyncStatus`.
4. Use structured concurrency in download iterations. We have recently restructured the control flow in the sync client for [Dart](https://github.com/powersync-ja/powersync.dart/pull/410), [JavaScript](https://github.com/powersync-ja/powersync-js/pull/951), Swift (in the rewrite) and [C#](https://github.com/powersync-ja/powersync-dotnet/pull/82). This adds that to Kotlin, a language that is very awesome for this: By using a [`ProducerScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/produce.html), we can syntactically group everything that generates sync events (the actual http stream, crud uploads, refreshed tokens, changed sync streams) into a single block that is cancelled when we stop consuming events. This untangles the control flow and makes client state much easier to reason about, a benefit for checkpoint requests which will add a new source of events.

AI use: I used Claude Code for a review and minor fixes.

### DIFF
--- a/common/src/commonTest/kotlin/powersync/sync/StreamingSyncClientTest.kt
+++ b/common/src/commonTest/kotlin/powersync/sync/StreamingSyncClientTest.kt
@@ -174,57 +174,6 @@ class StreamingSyncClientTest {
         }
 
     @Test
-    fun testStreamingSyncBasicFlow() =
-        runTest {
-            bucketStorage =
-                mock<BucketStorage> {
-                    everySuspend { getClientId() } returns "test-client-id"
-                }
-
-            streamingSyncClient =
-                StreamingSyncClient(
-                    status = status,
-                    bucketStorage = bucketStorage,
-                    connector = connector,
-                    uploadCrud = { },
-                    retryDelay = 10.milliseconds,
-                    logger = logger,
-                    params = JsonObject(emptyMap()),
-                    options =
-                        SyncOptions(
-                            clientConfiguration =
-                                SyncClientConfiguration.ExistingClient(
-                                    HttpClient(assertNoHttpEngine) {
-                                        configureSyncHttpClient()
-                                    },
-                                ),
-                        ),
-                    schema = Schema(),
-                    activeSubscriptions = MutableStateFlow(emptyList()),
-                )
-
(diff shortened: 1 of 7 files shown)


common/src/commonTest/kotlin/powersync/sync/StreamingSyncClientTest.kt
```diff
@@ -161,57 +173,6 @@ class StreamingSyncClientTest {
             }
         }
 
-    @Test
```
simolus3, line 164: This test is superseded by the much more complete `SyncIntegrationTest`, updating mocks to fix it doesn't really make sense.

Manrich121 (approved):

PR #373, opened 2026-08-20
Explicit checkpoint requests

simolus3:
This adds the `PowerSyncDatabase.requestCheckpoint()` method returning a `CheckpointRequest` instance which can be used to await an explicit checkpoint request. The API matches the existing Swift and JavaScript implementations. Tests are [ported from JavaScript](https://github.com/powersync-ja/powersync-js/blob/a875d48b8e99e0ce9999d3f8b305e56d2f554475/packages/node/tests/sync.test.ts#L218-L293).

As a simple demonstration, the multiplatform supabase todolist demo now has a refresh button in the app bar. Pressing it refreshes a checkpoint and waits for that to sync:

https://github.com/user-attachments/assets/8c476c83-fd02-4378-b5ff-732043bdbebd

AI use: I used Claude Code to update the demo and to review these changes.

### DIFF
--- a/common/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
+++ b/common/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
@@ -22,6 +22,8 @@ import com.powersync.db.internal.InternalTable
 import com.powersync.db.internal.PowerSyncVersion
 import com.powersync.db.schema.Schema
 import com.powersync.sync.CheckpointMode
+import com.powersync.sync.CheckpointRequest
+import com.powersync.sync.CheckpointRequestException
 import com.powersync.sync.StreamingSyncClient
 import com.powersync.sync.SyncOptions
 import com.powersync.sync.SyncStatus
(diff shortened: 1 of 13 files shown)


common/src/commonMain/kotlin/com/powersync/sync/CheckpointRequestException.kt
```diff
@@ -18,6 +18,10 @@ public open class CheckpointRequestException internal constructor(
 
     public class Disconnected internal constructor() : CheckpointRequestException("Cannot request checkpoints, sync client is disconnected")
 
-    public class NotEnabled internal constructor() :
+    public class Disabled internal constructor() :
```
simolus3, line 21: IMO, `Disabled` is a better name for the exception since we also have `Disconnected` (instead of `NotConnected`).

common/src/commonMain/kotlin/com/powersync/db/PowerSyncDatabaseImpl.kt
```diff
@@ -182,63 +189,58 @@ internal class PowerSyncDatabaseImpl(
...
-                // Get a global lock for checking mutex maps
-                val streamMutex = resource.group.syncMutex
-
-                // Poke the streaming mutex to see if another client is using it
-                var obtainedLock: HeldMutex? = null
-                try {
-                    // This call will throw if the lock is already held by this db client.
-                    // We should never reach that point since we disconnect before connecting.
-                    obtainedLock = streamMutex.tryAcquire(db)
+        val stream = createStream()
```
simolus3, line 194: This removes the `SupervisorJob`, which we don't really need since sync client methods already catch their own errors and retry. Injecting a custom job as a coroutine context with `launch` is something we're not really supposed to do either way.

Manrich121 (approved):

